PR #175, opened 2026-04-06
Add memory budget estimation API for RAID devices

szmyd:
## Summary

Adds static estimation methods to calculate worst-case memory usage **before** RAID device construction. This enables external services (e.g., gRPC handlers) to validate memory availability against cgroup limits before creating RAID devices, preventing OOM scenarios.

Key improvement: **Layered API** separates target-level (queue/thread) overhead from device-specific overhead, avoiding double-counting when devices are stacked.

## Memory Components Estimated

**Target Level** (shared across device stack):
- **ublksrv I/O buffers**: `(qdepth × max_io_size)` per queue (~64 MiB default)
- **Thread stacks**: 8 MiB per queue (OS default)
- **Target overhead**: ~4 KiB

**Device Level** (RAID-specific):
- **RAID-1**: SuperBlocks + worst-case bitmap (100% dirty pages, scales with volume size)

## API

```cpp
// Target-level: queue and thread overhead (reads runtime config from SISL options)
uint64_t ublkpp_tgt::estimate_queue_memory() noexcept;

// RAID-1: device-specific overhead (reads chunk_size from SISL options)
uint64_t Raid1Disk::estimate_device_overhead(uint64_t volume_size) noexcept;
```

**Why layered?** In stacked configurations (e.g., RAID-0 → RAID-1 → FSDisk), queue memory is allocated once at the target level and shared. Device overhead is specific to each RAID layer.

## Implementation Details

- Uses `sizeof(Bitmap::PageData)` with compile-time verification via `static_assert`
- Removed magic numbers in favor of `sizeof` for maintainability
- Runtime config (qdepth, max_io_size, nr_hw_queues, chunk_size) read from SISL options
- Only truly dynamic parameters (volume_size, num_disks) passed to methods
- Worst-case bitmap estimation for conservative safety
- All methods are `noexcept` static functions

## Recent Changes

**Use `sizeof` instead of magic numbers** (commits 4cdd316, 598041b):
- Replaced hardcoded `k_page_data_overhead` constant with `sizeof(raid1::Bitmap::PageData)`
- Added `static_assert(sizeof(raid1::Bitmap::PageData) == 24)` to catch structure changes at compile time
- Updated tests to use exact values (`EXPECT_EQ`) instead of ranges, now that size is guaranteed

This makes the code more robust: if PageData structure changes in the future, the build will fail with a clear error message prompting an update to the memory estimation.

## Test Coverage

- ✅ **6 unit tests** (5 RAID-1, 1 target)
- ✅ Exact value assertions for all memory calculations
- ✅ Volume size scaling (bitmap growth)
- ✅ Linear scaling validation (10x volume = ~10x memory)
- ✅ Edge cases (single disk, small/large volumes)
- ✅ All tests pass with address sanitizer

## Example Usage

```cpp
// In gRPC service handling RAID-1 creation requests
uint64_t total_memory = 
    ublkpp::ublkpp_tgt::estimate_queue_memory() +           // ~72 MiB (default config)
    ublkpp::Raid1Disk::estimate_device_overhead(volume_size); // ~4-40 MiB (1-10 TiB)

uint64_t available_memory = probe_cgroup_memory_available();
if (total_memory > available_memory) {
    return grpc::Status(grpc::StatusCode::RESOURCE_EXHAUSTED,
                       fmt::format("Insufficient memory: need {} MiB, have {} MiB",
                                 total_memory / (1024 * 1024),
                                 available_memory / (1024 * 1024)));
}

// Safe to create
auto raid1 = std::make_unique<ublkpp::Raid1Disk>(...);
```

## Known Limitations

1. **ublksrv buffer estimate**: Formula-based approximation (no API available to query actual size)
2. **Thread stack assumption**: Hardcoded 8 MiB (Linux default), may vary by OS
3. **Bitmap worst-case**: Assumes 100% dirty pages - may over-estimate by up to 10x in practice
4. **Static only**: Estimate is pre-construction; doesn't track runtime growth

These limitations are acceptable for OOM prevention use case, where conservative estimates are preferred.

## Test Results

All tests passing with exact value verification:
```
Target:  1/1 tests passed
RAID-1:  5/5 tests passed
```

**Example estimates** (default config: qdepth=128, max_io=512KiB, nr_queues=1):
- Queue overhead: ~72 MiB
- RAID-1 (512 MiB): 12,312 bytes (exact)
- RAID-1 (1 TiB): 4,227,072 bytes (exact)
- RAID-1 (10 TiB): 42,196,992 bytes (exact)


Ilyin-Ilya (commented):
Do we need all the docs?

PR #177, opened 2026-04-08
raid1: Fix route state race conditions (#176)

szmyd:
Fixes two races between __capture_route_state() and swap_device() identified in #176: raw MirrorDevice* pointers in RouteState could be invalidated by a concurrent swap, and the route/device reads were not atomic with each other, allowing a mismatched snapshot.

RouteState now holds shared_ptr<MirrorDevice> instead of raw pointers, keeping devices alive for the lifetime of any captured state. __capture_route_state() uses a double-read retry: copy both device shared_ptrs, read the route, then verify the pointers are unchanged — retrying if a swap landed in between. This is safe because __swap_device() always CAS-updates the route before swapping the pointer, so a stable pointer pair implies a consistent route. No mutex or sequence counter is needed and the reader path remains lock-free. The same fix is applied to __failover_read() and toggle_resync(), which previously bypassed the captured state and read _device_a/_device_b directly.


src/raid/raid1/raid1.cpp
```diff
@@ -628,46 +701,65 @@ io_result Raid1DiskImpl::__replicate(sub_cmd_t sub_cmd, auto&& func, uint64_t ad
...
         sub_cmd = shift_route(sub_cmd, route_size());
+    }
+
+    // Capture routing state atomically at function entry, or reuse passed state
+    RouteState const local_state = state ? RouteState{} : __capture_route_state(sub_cmd);
+    auto nested_retry = !!state;
+    if (!state) state = &local_state;
+
+    // Decode last_read from existing route bits, if nested call last_route is already set correctly.
+    if (retry && !nested_retry) {
```
Ilyin-Ilya: btw here we basically do 
retry && !!!old_state from line 717. Can be simplified as !old_state, but nevermind.

Ilyin-Ilya (approved):
Approved, fingers crossed.

Issue #180, opened 2026-04-09
RAID1 Concurrency Issues: Data Races and Memory Ordering Bugs


Ilyin-Ilya:
## Summary

Comprehensive concurrency analysis of the RAID1 subsystem identified **6 concurrency issues** ranging from critical data races to minor optimization opportunities.

**Critical:** 1  
**Important:** 3  
**Minor:** 2

---

### 🔴 Issue #1: `_resync_enabled` Data Race

**Location:** `raid1_impl.hpp:51`, `raid1.cpp:899`

**Problem:** `_resync_enabled` is a plain `bool` with unsynchronized reads/writes from multiple threads.

```cpp
bool _resync_enabled{true};  // ❌ Not atomic
```

**Concurrent Access:**
- **Writer:** User calls `toggle_resync(false)` from main thread (raid1.cpp:899)
- **Reader:** `__become_degraded()` checks `_resync_enabled` from I/O thread (raid1.cpp:588)

**Race Scenario:**
```
Thread 1 (I/O):   if (_resync_enabled) toggle_resync(true);
                     ↓ reads true
Thread 2 (API):   toggle_resync(false);
                     ↓ writes false
Thread 1:            ↓ launches resync (should have been skipped!)
```

**Impact:** Resync could launch when disabled, or fail to launch when enabled.

**Fix:**
```cpp
std::atomic<bool> _resync_enabled{true};

void toggle_resync(bool t) {
    _resync_enabled.store(t, std::memory_order_release);
    if (_resync_enabled.load(std::memory_order_acquire)) {
        // ...
    }
}
```

---

### 🟡 Issue #2: SuperBitmap Thread Safety

**Location:** `bitmap.hpp:63`, `super_bitmap.hpp`

**Problem:** SuperBitmap uses plain byte array with non-atomic bitwise operations.

```cpp
void set_bit(uint32_t page) noexcept {
    _bits[page / 8] |= (1U << (page % 8));  // ❌ Read-modify-write race
}
```

**Concurrent Access:**
- Resync thread: `clean_region()` → `clear_bit()`
- I/O thread: `dirty_region()` → `set_bit()`
- Multiple readers: `is_dirty()` → `test_bit()`

**Race Scenario:**
```
Thread 1: Read _bits[0] = 0b00000001
Thread 2: Read _bits[0] = 0b00000001
Thread 1: Write _bits[0] = 0b00000011 (set bit 1)
[...]

PR #184, opened 2026-04-10
raid1: Add UNAVAIL replica state for transient read failures

### LINKED ISSUE #183: Add read-unavailable state tracking for RAID1 devices
## Summary

Add a new `UNAVAIL` state for RAID1 devices that separates read failures from write failures, enabling auto-recovery from transient read issues without triggering unnecessary degradation and resync operations.

## Motivation

Currently, when a device experiences read failures, the only option is to degrade the device and trigger a full resync. However, read failures are often transient (cable flap, controller hiccup, temporary connectivity issues) while the underlying data remains valid. Forcing a full resync for transient issues is inefficient and puts unnecessary load on the system.

### New State: `UNAVAIL`

Add a fourth replica state alongside `CLEAN`, `SYNCING`, and `ERROR`:

- **CLEAN**: Device is healthy and fully active
- **SYNCING**: Device is degraded (missing data) but available for reads of valid regions  
- **ERROR**: Device is degraded AND has hardware/read issues
- **UNAVAIL**: Device has all data but is experiencing transient read failures

### Smart Load Balancing

When a device is in the `UNAVAIL` state:
- Route ~2% of read traffic to it (for auto-recovery attempts)
- Route 98% of traffic to the healthy replica
- Minimizes impact on failing device while allowing recovery detection

### Auto-Recovery

- On successful READ: automatically clear `UNAVAIL` flag and return device to full rotation
- No manual intervention required
- Seamless recovery from transient issues

### No Unnecessary Degradation

- Read failures set `UNAVAIL` flag but maintain `read_route::EITHER`
- Avoids triggering resync operations
- Write operations continue to both devices

## Benefits

1. **Reduced resync overhead**: Transient read failures don't trigger multi-GB resync operations
2. **Better availability**: Devices auto-recover from temporary issues
3. **Clear state visibility**: Operators can distinguish between "device has bad data" vs "device has temporary read issues"
4. **Improved I/O performance**: Failed devices receive minimal traffic during recovery period

## Implementation Details

The feature tracks read availability using an atomic flag per device and implements:
- Read failure detection in `__failover_read()`
- Smart load balancing with ~2% retry rate
- Auto-recovery on successful reads in `on_io_complete()`
- Proper state reporting in `replica_states()`

## Memory Overhead

Minimal: One `atomic_flag` per MirrorDevice (~1 byte)

## Testing

Comprehensive test coverage including:
- Read failure detection and UNAVAIL state transition
- Auto-recovery on successful reads
- Load balancing behavior (2% retry rate)
- State reporting accuracy
- Thread safety of atomic flag operations

szmyd:
Addresses #183, #186

## Summary

- Adds a fourth `replica_state` — **UNAVAIL** — to distinguish transiently unreachable devices from write-degraded ones
- Fixes a latent bug in `replica_states()` where a backup device could be reported CLEAN before the array had fully recovered
- Adds an idle-scoped periodic health monitor (`Raid1AvailProbeTask`) that detects and recovers from read failures while no user I/O is in flight
- Routes reads away from UNAVAIL devices (non-degraded mode only); recovery is handled by the idle probe
- Skips resync copy attempts when the dirty mirror is known unavailable, avoiding wasted I/O and log noise
- Simplifies `__failover_read` route selection logic

## Problem

RAID-1 previously tracked two degraded conditions with three states (CLEAN / SYNCING / ERROR), but only one of them reflected transient reachability:

| Condition | Before | After |
|---|---|---|
| Both devices healthy | CLEAN / CLEAN | CLEAN / CLEAN |
| Device missed writes (needs resync) | SYNCING or ERROR | SYNCING or ERROR |
| Device temporarily unreachable (read failure) | *silent* — treated as CLEAN | **UNAVAIL** |

A device that fails a read still has all of its data — it just can't be reached right now. Conflating this with write-degradation causes two problems:
1. Operators can't distinguish a flapping network path from actual data loss
2. RAID-1 had no mechanism to recover from a transient read failure without promoting it to a full write-degraded event

Additionally, when a device is UNAVAIL during resync, the old code would issue copy attempts that would fail on the write side every outer loop iteration — wasted I/O and error log spam until the device recovered.

### Latent bug in `replica_states()`

After `handle_internal()` successfully wrote a region to the backup device, `dirty_data_est()` reached zero. The old code then returned CLEAN for the backup device even while the route was still DEVA/DEVB — i.e., before `__become_clean()` had run and written the superblocks. This caused `wait_for_clean_state()` to return prematurely, leading to GMock expectation violations in `Raid1.CleanBitmap`.

### UNAVAIL state (`replica_state::UNAVAIL = 3`)

- `__failover_read()` sets `dev->unavail` on any failed read attempt (before retrying the other device)
- `on_io_complete()` clears `dev->unavail` on any successful user-facing READ completion (auto-recovery — no operator intervention required)
- `replica_states()` reports UNAVAIL when `route == EITHER` but the device has `unavail` set — meaning the array is otherwise healthy but that device has recent read failures

### Read routing for UNAVAIL devices

In non-degraded mode, if the load-balancer selects an UNAVAIL device, the read is routed to the other device instead. Recovery is detected passively via `on_io_complete` (on the next successful read) or actively via the idle probe. This check is skipped in degraded mode, where `unavail` is set structurally by `__become_degraded` and routing is already handled by the existing degraded-mode logic.

### Idle-scoped periodic health monitor (`Raid1AvailProbeTask`)

When the queue goes idle (no user I/O in flight), a background thread periodically reads a page from each mirror device. This handles recovery detection when the array is idle and no user reads are flowing to trigger the `on_io_complete` auto-recovery path. The probe runs only while idle and is stopped on idle exit.

### Resync unavail guard

At the top of the outer resync copy loop, if the dirty mirror is flagged UNAVAIL, the task skips straight to `__yield(unavail_delay)` without issuing any I/O. Previously, each outer iteration would attempt at least one `__copy_region` that would fail on the write side before yielding.

### `replica_states()` refactor and bug fix

Extracted a `get_state` helper lambda that correctly handles all combinations of route and unavail flag:

```cpp
// BEFORE: backup device with empty bitmap -> CLEAN, even if route != EITHER
if (!is_active && sync_bytes > 0) {
    return is_unavail ? replica_state::ERROR : replica_state::SYNCING;
}

// AFTER: backup device stays SYNCING until __become_clean() restores the route
if (!is_active && (sync_bytes > 0 || state.route != read_route::EITHER)) {
    return is_unavail ? replica_state::ERROR : replica_state::SYNCING;
}
```

### `__failover_read` simplification

- `need_to_test` bool replaced with a direct `route != state->route` comparison
- Round-robin else block collapsed to a ternary
- Dead `if (!retry)` guard removed from the read-failure else branch (unreachable: `retry==true` always exits via `res || retry` above)

## Hot path overhead

On the happy path (both devices healthy, successful read) this PR adds:

- **`__failover_read`**: one `atomic_flag::test(acquire)` on `unavail`, gated by `!is_degraded`
- **`on_io_complete`**: one `atomic_flag::test(acquire)` on `unavail`, gated by `UBLK_IO_OP_READ && res>=0 && !is_internal`

On x86-64, `atomic_flag::test(acquire)` is a plain `movzx` load with no lock prefix or fence. `unavail` lives inside `MirrorDevice` alongside the `disk` shared_ptr already dereferenced in `__route_to_device`, so the cache line is already hot. Overhead is not measurable in practice.

## Impact

- **Operators** get a clear signal when a device is transiently unreachable vs. write-degraded — no false alarms, no unnecessary resync
- **Auto-recovery** happens silently on the next successful read or idle probe; no manual intervention needed for transient failures
- **Data safety**: UNAVAIL never skips writes — both devices still receive every write. Only reads are rerouted
- **Resync efficiency**: No wasted I/O or log spam while waiting for an unavailable mirror to recover
- **`replica_states()` fix**: `wait_for_clean_state()` now correctly blocks until `__become_clean()` has committed superblocks to both devices

## Test Plan

- [x] `ReadFailureSetsUnavail` — single read failure sets UNAVAIL, route stays EITHER, bytes_to_sync stays 0
- [x] `SuccessfulReadClearsUnavail` — successful read on previously-UNAVAIL device clears the flag (auto-recovery)
- [x] `ReadFailureDoesNotDegrade` — both devices UNAVAIL still allows writes; route remains EITHER
- [x] `WriteDegradedShowsError` — write failure still produces ERROR (not UNAVAIL), distinguishing the two conditions
- [x] `SyncIoTracksReadFailures` — sync I/O path also sets UNAVAIL on failure
- [x] `Raid1.CleanBitmap` — previously failing; now passes with `replica_states()` fix
- [x] All 5 test suites pass (100%) on remote Debug, TSAN, and ASAN builds

### DIFF
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -751,21 +751,15 @@ io_result Raid1DiskImpl::__failover_read(sub_cmd_t sub_cmd, auto&& func, uint64_
 
     // Pick a device to read from (load-balancer)
     auto route = read_route::DEVA;
-    auto need_to_test{false};
-    if (state->is_degraded && (!retry && state->backup_dev->unavail.test(std::memory_order_acquire))) {
+    if (state->is_degraded && !retry && state->backup_dev->unavail.test(std::memory_order_acquire)) {
         route = state->route;
     } else {
-        if (read_route::DEVB == last_read) {
-            if (read_route::DEVB == state->route) need_to_test = true;
-        } else {
-            route = read_route::DEVB;
-            if (read_route::DEVA == state->route) need_to_test = true;
-        }
+        route = (read_route::DEVB == last_read) ? read_route::DEVA : read_route::DEVB;
     }
 
-    // If we are degraded and the load-balancer wants to use the dirty disk, check it for dirty bits first
-    // and if all true; then  use the current active route from the captured state
-    if (state->is_degraded && need_to_test && _dirty_bitmap->is_dirty(addr, len)) route = state->route;
+    // In degraded mode, if the load-balancer picked the backup (dirty) device, verify the region is
+    // clean before using it — otherwise fall back to the active route
+    if (state->is_degraded && route != state->route && _dirty_bitmap->is_dirty(addr, len)) route = state->route;
 
     // We've already attempted this device...we don't want to re-attempt
     if (retry && (last_read == route)) return std::unexpected(std::make_error_condition(std::errc::io_error));
@@ -790,11 +784,9 @@ io_result Raid1DiskImpl::__failover_read(sub_cmd_t sub_cmd, auto&& func, uint64_
     if (auto res = func(*chosen_dev->disk, chosen_sub); res || retry) {
         return res;
     } else {
-        // On read failure (not retry), mark device as unavailable
-        if (!retry) {
-            if (!chosen_dev->unavail.test_and_set(std::memory_order_acquire)) {
-                RLOGW("Device marked unavailable due to read failure: {}", *chosen_dev->disk)
-            }
+        // On read failure mark device as unavailable before retrying the other side
+        if (!chosen_dev->unavail.test_and_set(std::memory_order_acquire)) {
+            RLOGW("Device marked unavailable due to read failure: {}", *chosen_dev->disk)
         }
     }
 
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -941,7 +941,7 @@ void Raid1DiskImpl::on_io_complete(ublk_io_data const* data, sub_cmd_t sub_cmd, 
     }
 }
 
-void Raid1DiskImpl::idle_transition(ublksrv_queue const*, bool enter) {
+void Raid1DiskImpl::idle_transition(ublksrv_queue const*, bool enter) noexcept {
     if (!enter) {
         _idle_probe_a.stop();
         _idle_probe_b.stop();
@@ -951,7 +951,10 @@ void Raid1DiskImpl::idle_transition(ublksrv_queue const*, bool enter) {
     auto const state = __capture_route_state();
     if (state.is_degraded) return; // Resync task handles avail probing in degraded mode
 
-    // Immediate probe: clear UNAVAIL on any device that has recovered (edge trigger, no delay)
+    // Immediate probe: clear UNAVAIL on any device that has recovered (edge trigger, no delay).
+    // TOCTOU note: on_io_complete() may clear unavail between the test() and probe_mirror().
+    // If so, probe_mirror() still succeeds (device reachable) and the log fires spuriously.
+    // This is benign — the probe is idempotent and the worst case is one redundant log entry.
     auto const immediate_probe = [&](std::shared_ptr< MirrorDevice > const& mirror) {
         if (!mirror->unavail.test(std::memory_order_acquire)) return;
         if (probe_mirror(*mirror, reserved_size)) RLOGD("Idle probe: device recovered: {}", *mirror->disk)
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -969,7 +969,7 @@ void Raid1DiskImpl::idle_transition(ublksrv_queue const*, bool enter) {
     if (state.is_degraded) return; // Resync task handles avail probing in degraded mode
 
     // Immediate probe: clear UNAVAIL on any device that has recovered (edge trigger, no delay)
-    static thread_local alignas(raid1::k_page_size) uint8_t probe_buf[raid1::k_page_size];
+    alignas(raid1::k_page_size) static thread_local uint8_t probe_buf[raid1::k_page_size];
     auto const immediate_probe = [&](std::shared_ptr< MirrorDevice > const& mirror) {
         if (!mirror->unavail.test(std::memory_order_acquire)) return;
         auto iov = iovec{.iov_base = probe_buf, .iov_len = raid1::k_page_size};
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -770,8 +770,10 @@ io_result Raid1DiskImpl::__failover_read(sub_cmd_t sub_cmd, auto&& func, uint64_
     // We've already attempted this device...we don't want to re-attempt
     if (retry && (last_read == route)) return std::unexpected(std::make_error_condition(std::errc::io_error));
 
-    // Route away from unavail devices; recovery is handled by the idle probe
-    if (__route_to_device(*state, route).device->unavail.test(std::memory_order_acquire)) {
+    // Route away from unavail devices; recovery is handled by the idle probe.
+    // In degraded mode unavail is set on the backup device as part of degradation — routing is
+    // already handled by the is_degraded block above, so skip this check there.
+    if (!state->is_degraded && __route_to_device(*state, route).device->unavail.test(std::memory_order_acquire)) {
         route = (route == read_route::DEVA) ? read_route::DEVB : read_route::DEVA;
         RLOGD("Skipping unavail device, routing to alternate")
     }
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -770,22 +770,8 @@ io_result Raid1DiskImpl::__failover_read(sub_cmd_t sub_cmd, auto&& func, uint64_
     // We've already attempted this device...we don't want to re-attempt
     if (retry && (last_read == route)) return std::unexpected(std::make_error_condition(std::errc::io_error));
 
-    // Smart load balancing: For unavail devices, only retry ~2% of the time
-    // This allows auto-recovery while minimizing impact on failing devices
-    bool const retry_unavail = [&]() -> bool {
-        // Check if chosen device is unavailable
-        auto const& dev = __route_to_device(*state, route).device;
-        bool const chosen_unavail = dev->unavail.test(std::memory_order_acquire);
-
-        if (!chosen_unavail) return true; // Device available, always use it
-
-        // Device unavailable: retry only ~2% of the time for auto-recovery
-        static thread_local std::atomic< uint32_t > read_counter{0};
-        return (read_counter.fetch_add(1, std::memory_order_relaxed) % 50) == 0;
-    }();
-
-    if (!retry_unavail) {
-        // Skip this unavail device, try the other one
+    // Route away from unavail devices; recovery is handled by the idle probe
+    if (__route_to_device(*state, route).device->unavail.test(std::memory_order_acquire)) {
         route = (route == read_route::DEVA) ? read_route::DEVB : read_route::DEVA;
         RLOGD("Skipping unavail device, routing to alternate")
     }
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -958,6 +958,34 @@ void Raid1DiskImpl::on_io_complete(ublk_io_data const* data, sub_cmd_t sub_cmd, 
     }
 }
 
+void Raid1DiskImpl::idle_transition(ublksrv_queue const*, bool enter) {
+    if (!enter) {
+        _idle_probe_a.stop();
+        _idle_probe_b.stop();
+        return;
+    }
+
+    auto const state = __capture_route_state();
+    if (state.is_degraded) return; // Resync task handles avail probing in degraded mode
+
+    // Immediate probe: clear UNAVAIL on any device that has recovered (edge trigger, no delay)
+    static thread_local alignas(raid1::k_page_size) uint8_t probe_buf[raid1::k_page_size];
+    auto const immediate_probe = [&](std::shared_ptr< MirrorDevice > const& mirror) {
+        if (!mirror->unavail.test(std::memory_order_acquire)) return;
+        auto iov = iovec{.iov_base = probe_buf, .iov_len = raid1::k_page_size};
+        if (auto res = mirror->disk->sync_iov(UBLK_IO_OP_READ, &iov, 1, reserved_size); res) {
+            mirror->unavail.clear(std::memory_order_release);
+            RLOGD("Idle probe: device recovered: {}", *mirror->disk)
+        }
+    };
+    immediate_probe(state.active_dev);
+    immediate_probe(state.backup_dev);
+
+    // Periodic probe: both directions (recovery + new failures), stopped when idle exits
+    _idle_probe_a.launch(state.active_dev, reserved_size);
+    _idle_probe_b.launch(state.backup_dev, reserved_size);
+}
+
 void Raid1DiskImpl::toggle_resync(bool t) {
     _resync_enabled = t;
     if (_resync_enabled) {
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -503,23 +503,38 @@ raid1::array_state Raid1DiskImpl::replica_states() const noexcept {
     auto const sz_to_sync = _dirty_bitmap->dirty_data_est();
     auto const state = __capture_route_state();
 
+    // Helper: compute state for a device
+    auto const get_state = [&state](MirrorDevice const* dev, bool is_active, uint64_t sync_bytes) -> replica_state {
+        bool const is_unavail = dev->unavail.test(std::memory_order_acquire);
+
+        if (!is_active && (sync_bytes > 0 || state.route != read_route::EITHER)) {
+            // Device is degraded (doesn't have all data or route not yet restored)
+            return is_unavail ? replica_state::ERROR : replica_state::SYNCING;
+        }
+
+        if (is_unavail && state.route == read_route::EITHER) {
+            // Healthy array, but device has read failures
+            return replica_state::UNAVAIL; // Has data, can't reach it
+        }
+
+        return replica_state::CLEAN;
+    };
+
     switch (state.route) {
-    case read_route::DEVA:
-        return raid1::array_state{.device_a = replica_state::CLEAN,
-                                  .device_b = state.backup_dev->unavail.test(std::memory_order_acquire)
-                                      ? replica_state::ERROR
-                                      : replica_state::SYNCING,
-                                  .bytes_to_sync = sz_to_sync};
-    case read_route::DEVB:
-        return raid1::array_state{.device_a = state.backup_dev->unavail.test(std::memory_order_acquire)
-                                      ? replica_state::ERROR
-                                      : replica_state::SYNCING,
-                                  .device_b = replica_state::CLEAN,
-                                  .bytes_to_sync = sz_to_sync};
-    case read_route::EITHER:
+    case read_route::DEVA: // Device B is write-degraded
+        return {.device_a = get_state(state.active_dev.get(), true, sz_to_sync),
+                .device_b = get_state(state.backup_dev.get(), false, sz_to_sync),
+                .bytes_to_sync = sz_to_sync};
+    case read_route::DEVB: // Device A is write-degraded
+        return {.device_a = get_state(state.backup_dev.get(), false, sz_to_sync),
+                .device_b = get_state(state.active_dev.get(), true, sz_to_sync),
+                .bytes_to_sync = sz_to_sync};
+    case read_route::EITHER: // Healthy array
     default:
-        return raid1::array_state{
-            .device_a = replica_state::CLEAN, .device_b = replica_state::CLEAN, .bytes_to_sync = 0};
+        // For EITHER route: active_dev==device_a, backup_dev==device_b by convention
+        return {.device_a = get_state(state.active_dev.get(), true, 0),
+                .device_b = get_state(state.backup_dev.get(), true, 0),
+                .bytes_to_sync = 0};
     }
 }
 
@@ -755,6 +770,26 @@ io_result Raid1DiskImpl::__failover_read(sub_cmd_t sub_cmd, auto&& func, uint64_
     // We've already attempted this device...we don't want to re-attempt
     if (retry && (last_read == route)) return std::unexpected(std::make_error_condition(std::errc::io_error));
 
+    // Smart load balancing: For unavail devices, only retry ~2% of the time
+    // This allows auto-recovery while minimizing impact on failing devices
+    bool const retry_unavail = [&]() -> bool {
+        // Check if chosen device is unavailable
+        auto const& dev = __route_to_device(*state, route).device;
+        bool const chosen_unavail = dev->unavail.test(std::memory_order_acquire);
+
+        if (!chosen_unavail) return true; // Device available, always use it
+
+        // Device unavailable: retry only ~2% of the time for auto-recovery
+        static thread_local std::atomic< uint32_t > read_counter{0};
+        return (read_counter.fetch_add(1, std::memory_order_relaxed) % 50) == 0;
+    }();
+
+    if (!retry_unavail) {
+        // Skip this unavail device, try the other one
+        route = (route == read_route::DEVA) ? read_route::DEVB : read_route::DEVA;
+        RLOGD("Skipping unavail device, routing to alternate")
+    }
+
     // Move load-balancer forward, this is optimistic, doesn't need to be atomic
     last_read = route;
 
@@ -764,7 +799,16 @@ io_result Raid1DiskImpl::__failover_read(sub_cmd_t sub_cmd, auto&& func, uint64_
     // Add RETRIED flag if this is a retry attempt
     auto const chosen_sub = retry ? set_flags(chosen_sub_base, sub_cmd_flags::RETRIED) : chosen_sub_base;
 
-    if (auto res = func(*chosen_dev->disk, chosen_sub); res || retry) { return res; }
+    if (auto res = func(*chosen_dev->disk, chosen_sub); res || retry) {
+        return res;
+    } else {
+        // On read failure (not retry), mark device as unavailable
+        if (!retry) {
+            if (!chosen_dev->unavail.test_and_set(std::memory_order_acquire)) {
+                RLOGW("Device marked unavailable due to read failure: {}", *chosen_dev->disk)
+            }
+        }
+    }
 
     // Otherwise fail over the device and attempt the READ again marking this a retry
     sub_cmd = set_flags(sub_cmd, sub_cmd_flags::RETRIED);
@@ -893,6 +937,16 @@ void Raid1DiskImpl::on_io_complete(ublk_io_data const* data, sub_cmd_t sub_cmd, 
     // Pass completion notification to the underlying device for its metrics
     // Map orig_route to physical device accounting for potential swap
     __route_to_device(state, orig_route).device->disk->on_io_complete(data, sub_cmd, res);
+
+    // Auto-recovery: Clear unavail flag on successful READ completions (user I/O only, not internal/resync)
+    if (UBLK_IO_OP_READ == ublksrv_get_op(data->iod) && res >= 0 && !is_internal(sub_cmd)) {
+        auto const& device = __route_to_device(state, orig_route).device;
+
+        if (device->unavail.test(std::memory_order_acquire)) {
+            device->unavail.clear(std::memory_order_release);
+            RLOGD("Device auto-recovered from read failure: {}", *device->disk)
+        }
+    }
 
     // Decrement outstanding write counter for writes (not reads), but only if it was a
     // success.
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -9,6 +9,7 @@
 #include <sisl/options/options.h>
 
 #include "bitmap.hpp"
+#include "raid1_avail_probe.hpp"
 #include "raid1_impl.hpp"
 #include "raid1_resync_task.hpp"
 #include "lib/logging.hpp"
@@ -487,9 +488,11 @@ std::shared_ptr< UblkDisk > Raid1DiskImpl::swap_device(std::string const& outgoi
     }
 
     // Atomically swap the device or fail; fail if swapping sole active device
-    if (__swap_device(outgoing_device_id, incoming_mirror, state.route) && _raid_metrics) {
-        // Record successful device swap
-        _raid_metrics->record_device_swap();
+    if (__swap_device(outgoing_device_id, incoming_mirror, state.route)) {
+        if (_raid_metrics) _raid_metrics->record_device_swap();
+        // Stop stale probes — they hold a shared_ptr to the outgoing MirrorDevice
+        _idle_probe_a.stop();
+        _idle_probe_b.stop();
     }
 
     // Now set back to IDLE state and kick a resync task off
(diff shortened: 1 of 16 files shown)


Ilyin-Ilya:
## Code Review Summary

### Assessment
- 🔴 Critical Issues: 0
- 🟡 Important Issues: 2
- 🟢 Minor Issues: 3
- ✅ Excellent design and implementation

---

## Strengths

### Architectural Excellence
- **Clear state separation**: UNAVAIL vs ERROR vs SYNCING distinguishes transient network failures from write-degradation
- **Auto-recovery**: Silent recovery on successful read eliminates operator toil
- **Minimal hot-path overhead**: One atomic test on already-hot cache line (unmeasurable on x86-64)
- **State machine correctness**: Idle probe lifecycle properly managed (stop on swap, stop on idle exit)

### Code Quality
- **Excellent refactoring**: `replica_states()` helper lambda eliminates duplication, fixes latent bug
- **Comprehensive tests**: 389 lines of new tests covering all edge cases
- **Clear documentation**: PR description explains problem, solution, impact, and overhead analysis

### Bug Fixes
1. **`replica_states()` latent bug** - Backup device could report CLEAN before `__become_clean()` → fixed by checking `route != EITHER`
2. **Idle probe UAF** - swap_device now stops probes holding outgoing MirrorDevice

---

## Issues Found

### 🟡 Important #1: Potential Race in `idle_transition()`

**Location:** `raid1.cpp:196-217`

**Issue:** Read-then-use race on `mirror->unavail`

```cpp
auto const immediate_probe = [&](std::shared_ptr<MirrorDevice> const& mirror) {
    if (!mirror->unavail.test(std::memory_order_acquire)) return;  // ← Read
    if (probe_mirror(*mirror, reserved_size))                      // ← Use (gap!)
        RLOGD("Idle probe: device recovered: {}", *mirror->disk)
};
```

**Race scenario:**
```
Time    Thread 1 (idle_transition)       Thread 2 (on_io_complete)
T0      if (!unavail.test())             
          ↓ reads TRUE (unavail set)
T1                                        unavail.clear()  // Auto-recovery
T2      probe_mirror(*mirror, ...)       
          ↓ Redundant probe (benign)
          ↓ Logs "device recovered"      [Already recovered by T2!]
```

**Impact:** Benign - redundant probe + misleading log. Not a data race (atomic operations), but TOCTOU.

**Fix:** Either accept as benign (probe is idempotent), or use test-and-clear:
```cpp
if (mirror->unavail.test(std::memory_order_acquire)) {
    if (probe_mirror(*mirror, reserved_size)) {
        // Only log if probe actually cleared the flag
        RLOGD("Idle probe: device recovered: {}", *mirror->disk)
    }
}
```

---

### 🟡 Important #2: Resync Starvation on Oscillating Device

**Location:** `raid1_resync_task.cpp:422-429`

**Issue:** If dirty mirror oscillates (available → unavailable → available), resync never makes progress.

```cpp
while (0 < nr_pages) {
    if (dirty_mirror->unavail.test(...)) {
        __yield(unavail_delay, avail_delay);
        probe_mirror(*dirty_mirror, _offset);
        nr_pages = _dirty_bitmap->dirty_pages();  // ← Recalculates
        continue;  // ← Outer loop restarts, never copies data
    }
    // ... copy logic ...
}
```

**Scenario:**
1. Resync starts, 100 dirty pages
2. Device fails, outer loop sleeps 5s (unavail_delay)
3. Probe succeeds, continues outer loop
4. Device fails again before copying data
5. Infinite loop (no progress)

**Impact:** Moderate - resync starvation on flapping device. Logs will show repeated probe attempts.

**Fix:** Add retry limit or progress tracking:
```cpp
uint32_t consecutive_unavail_loops = 0;
while (0 < nr_pages) {
    if (dirty_mirror->unavail.test(...)) {
        if (++consecutive_unavail_loops > 10) {
            RLOGW("Dirty mirror repeatedly unavailable, aborting resync");
            break;
        }
        __yield(unavail_delay, avail_delay);
        probe_mirror(*dirty_mirror, _offset);
        nr_pages = _dirty_bitmap->dirty_pages();
        continue;
    }
    consecutive_unavail_loops = 0;  // Reset on successful iteration
    // ... copy logic ...
}
```

---

### 🟢 Minor #1: Missing `noexcept` on `idle_transition()`

**Location:** `raid1_impl.hpp:342`

```cpp
void idle_transition(ublksrv_queue const* q, bool enter) override;
```

Should be `noexcept` for consistency with UblkDisk virtual interface pattern. The implementation doesn't throw.

---

### 🟢 Minor #2: Redundant `is_unavail` Read in `get_state` Lambda

**Location:** `raid1.cpp:72`

```cpp
auto const get_state = [&state](MirrorDevice const* dev, bool is_active, uint64_t sync_bytes) {
    bool const is_unavail = dev->unavail.test(std::memory_order_acquire);  // ← Read once
    
    if (!is_active && (sync_bytes > 0 || state.route != read_route::EITHER)) {
        return is_unavail ? replica_state::ERROR : replica_state::SYNCING;
    }
    
    if (is_unavail && state.route == read_route::EITHER) {  // ← Used twice
        return replica_state::UNAVAIL;
    }
    
    return replica_state::CLEAN;
};
```

**Not a bug** - capturing `is_unavail` ensures both uses see same value. This is correct; just noting it's intentional.

---

### 🟢 Minor #3: `probe_mirror()` Return Value Semantics

**Location:** `raid1_avail_probe.cpp:241-250`

```cpp
bool probe_mirror(MirrorDevice& mirror, uint64_t reserved_size) noexcept {
    alignas(k_page_size) uint8_t probe_buf[k_page_size];
    auto iov = iovec{.iov_base = probe_buf, .iov_len = k_page_size};
    if (auto res = mirror.disk->sync_iov(UBLK_IO_OP_READ, &iov, 1, reserved_size); res) {
        mirror.unavail.clear(std::memory_order_release);
        return true;   // ← "Device is available"
    }
    mirror.unavail.test_and_set(std::memory_order_acquire);
    return false;  // ← "Device failed probe"
}
```

**Observation:** Function name `probe_mirror` and return semantics are slightly ambiguous:
- Returns `true` = "device is NOW available" (probe succeeded, flag cleared)
- Returns `false` = "device failed probe" (flag set/remains set)

**Suggestion:** Add doc comment clarifying semantics, or rename to `probe_and_clear_unavail()` for clarity.

---

## Code Conventions

✅ **Style**: 4-space indent, 120-char lines, left pointer alignment - all followed  
✅ **Naming**: PascalCase classes, snake_case functions, proper enum naming  
✅ **Error handling**: Uses `io_result` correctly  
✅ **Logging**: Proper SISL macros (`RLOGD`, `RLOGW`, `RLOGI`)  
✅ **Memory safety**: Aligned buffers, proper cleanup  
✅ **Tests**: Comprehensive coverage (5 new tests + 1 regression fix)

---

## Testing

✅ All 5 new tests pass (100%)  
✅ Fixes previously failing `Raid1.CleanBitmap` test  
✅ Tests cover:
- Single read failure → UNAVAIL state
- Auto-recovery on successful read
- Read failure doesn't trigger degradation
- Write degradation → ERROR (not UNAVAIL)
- Sync I/O path tracks failures
- Resync unblocks via probe_mirror
- Swap stops idle probes (UAF prevention)

---

## Concurrency Analysis

**Thread safety:**
- ✅ `unavail` is `atomic_flag` - all operations are atomic
- ✅ Idle probe uses `std::jthread` with `stop_token` - clean shutdown
- ✅ No new data races introduced
- ✅ Properly coordinates with existing resync state machine

**Memory ordering:**
- ✅ `test()` uses acquire, `clear()` uses release (correct publish/subscribe)
- ✅ `test_and_set()` uses acquire (conservative, safe)

**Edge cases handled:**
- ✅ Swap stops stale probes (prevents UAF)
- ✅ Probe skips degraded arrays (resync handles it)
- ✅ Route check skips UNAVAIL in degraded mode (already handled)

---

## Recommendations

### Must Fix Before Merge
None - code is production-ready as-is.

### Should Fix (Low Priority)
1. Add retry limit to resync unavail loop (prevents starvation on flapping device)
2. Make `idle_transition()` noexcept for consistency

### Nice to Have
1. Add doc comment to `probe_mirror()` clarifying return semantics
2. Consider logging level for redundant recovery (RLOGD → RLOGV)

---

## Performance Impact

**Hot path overhead (healthy array):**
- `__failover_read`: +1 atomic test (gated by `!is_degraded`)
- `on_io_complete`: +1 atomic test (gated by `READ && success && !internal`)

On x86-64: `movzx` load, no lock prefix, cache line already hot → **unmeasurable**.

**Idle path overhead:**
- Periodic probe thread (5s default) - minimal CPU, I/O only when idle

**Degraded path:**
- Resync now skips failed writes entirely → **reduces** I/O waste and log spam

Overall: **Net improvement** in efficiency.

---

## Conclusion

**Excellent PR.** Solves a real operational problem (transient read failures → false degradation alarms) with a clean architectural solution. Code quality is high, tests are comprehensive, and performance impact is negligible.

The two important issues (TOCTOU in immediate_probe, resync starvation) are edge cases that don't block merge but should be tracked for follow-up.

**Recommendation:** ✅ **Approve with minor follow-ups**

---

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

_Note: This is a review comment only. Approval/rejection will be done manually._

src/raid/raid1/raid1.cpp
```diff
@@ -894,6 +921,16 @@ void Raid1DiskImpl::on_io_complete(ublk_io_data const* data, sub_cmd_t sub_cmd,
     // Map orig_route to physical device accounting for potential swap
     __route_to_device(state, orig_route).device->disk->on_io_complete(data, sub_cmd, res);
 
+    // Auto-recovery: Clear unavail flag on successful READ completions (user I/O only, not internal/resync)
+    if (UBLK_IO_OP_READ == ublksrv_get_op(data->iod) && res >= 0 && !is_internal(sub_cmd)) {
+        auto const& device = __route_to_device(state, orig_route).device;
+
+        if (device->unavail.test(std::memory_order_acquire)) {
+            device->unavail.clear(std::memory_order_release);
+            RLOGD("Device auto-recovered from read failure: {}", *device->disk)
```
Ilyin-Ilya, line 930: maybe RDLOGI? this op seems important

src/raid/raid1/raid1_avail_probe.hpp
```diff
@@ -0,0 +1,24 @@
...
+// sets unavail on failure. Returns true if device is available.
+bool probe_mirror(MirrorDevice& mirror, uint64_t reserved_size) noexcept;
+
+class Raid1AvailProbeTask {
+    std::jthread _probe;
+
+public:
+    void launch(std::shared_ptr< MirrorDevice > mirror, uint64_t reserved_size);
+    void stop() noexcept { _probe.request_stop(); }
+    bool is_running() const noexcept { return _probe.joinable(); }
```
Ilyin-Ilya: we don't use it.

Ilyin-Ilya:
## Additional Review: Potential Race Conditions in Idle Probe

After deeper analysis of the idle probe mechanism, I've identified several potential concurrency issues:

---

## 🔴 Critical Issue #1: Concurrent `idle_transition()` Race

**Location:** `raid1.cpp:196-217`

**Problem:** Multiple threads can call `idle_transition()` concurrently with no synchronization.

```cpp
void idle_transition(ublksrv_queue const*, bool enter) {
    if (!enter) {
        _idle_probe_a.stop();  // ← No lock
        _idle_probe_b.stop();
        return;
    }
    
    // ... immediate probe ...
    
    _idle_probe_a.launch(state.active_dev, reserved_size);  // ← No lock
    _idle_probe_b.launch(state.backup_dev, reserved_size);
}
```

**Race scenario:**
```
Thread 1 (Q1 entering idle):        Thread 2 (I/O arrives, exiting idle):
  T0: Captures state                
  T1: Runs immediate_probe          
  T2: About to launch probes...     
                                    T3: _idle_probe_a.stop()
                                    T4: _idle_probe_b.stop()
  T5: _idle_probe_a.launch(...)  ← Launches AFTER stop!
  T6: _idle_probe_b.launch(...)
```

**Impact:** Probes running during active I/O, exactly what we tried to avoid.

**Suggested fix:**
```cpp
std::atomic<bool> _idle_probes_active{false};

void idle_transition(ublksrv_queue const*, bool enter) {
    if (!enter) {
        bool expected = true;
        if (_idle_probes_active.compare_exchange_strong(expected, false)) {
            _idle_probe_a.stop();
            _idle_probe_b.stop();
        }
        return;
    }
    
    bool expected = false;
    if (!_idle_probes_active.compare_exchange_strong(expected, true)) {
        return;  // Already launched
    }
    
    // ... launch probes ...
}
```

---

## 🔴 Critical Issue #2: `launch()` Not Thread-Safe

**Location:** `raid1_avail_probe.hpp:296`

**Problem:** `launch()` assigns to `_probe` member without synchronization.

```cpp
void launch(std::shared_ptr<MirrorDevice> mirror, uint64_t rs) {
    _probe = std::jthread([...] { ... });  // ← Data race on member
}
```

**Race scenario:**
```
Thread 1: _idle_probe_a.launch(...)
            → _probe = jthread(...)  // Destroys old jthread

Thread 2: _idle_probe_a.launch(...)  (concurrent)
            → _probe = jthread(...)  // ALSO destroys old jthread

Result: Data race on _probe member
```

**Impact:** Undefined behavior (data race on non-atomic member).

**Suggested fix:**
```cpp
class Raid1AvailProbeTask {
    std::jthread _probe;
    std::mutex _probe_mutex;  // Add mutex

public:
    void launch(std::shared_ptr<MirrorDevice> mirror, uint64_t rs) {
        std::scoped_lock lock(_probe_mutex);
        _probe = std::jthread([...] { ... });
    }
    
    void stop() noexcept {
        std::scoped_lock lock(_probe_mutex);
        _probe.request_stop();
    }
};
```

---

## 🟡 Important Issue #3: Multi-Queue Idle Tracking

**Location:** `raid1.cpp:196`

**Problem:** Global idle state for per-queue transitions.

If system has multiple queues (e.g., `nr_hw_queues > 1`), `idle_transition()` receives per-queue notifications but tracks global state:

```
Q1: Has I/O (active)
Q2: Goes idle → idle_transition(q2, true)
      ↓ Launches probes (but Q1 still active!)
      
Q1: Finishes I/O → idle_transition(q1, true)
      ↓ Launches probes again (overwrites)
```

**Impact:** Probes may run while some queues active, or not run while some queues idle.

**Suggested fix:**
```cpp
std::atomic<uint32_t> _active_queue_count{0};

void idle_transition(ublksrv_queue const* q, bool enter) {
    if (!enter) {
        // Queue becoming active
        if (_active_queue_count.fetch_add(1, std::memory_order_acq_rel) == 0) {
            // First queue active - stop probes
            _idle_probe_a.stop();
            _idle_probe_b.stop();
        }
        return;
    }
    
    // Queue becoming idle
    if (_active_queue_count.fetch_sub(1, std::memory_order_acq_rel) == 1) {
        // Last queue idle - start probes
        // ... launch probes ...
    }
}
```

---

## 🟡 Important Issue #4: `launch()` Blocks on Slow Join

**Location:** `raid1_avail_probe.cpp:252`

**Problem:** `stop()` only requests stop; doesn't wait for thread exit.

```cpp
void stop() noexcept { _probe.request_stop(); }

void launch(...) {
    _probe = std::jthread(...);  // ← Destructor joins old thread
}
```

**Scenario:**
```
T0: Probe thread sleeping in cv.wait_for() (up to 5s)
T1: exit idle → stop() (just sets stop_token)
T2: enter idle → launch()
      ↓ Assignment destroys old jthread
      ↓ Blocks waiting for join (up to 5s!)
```

**Impact:** `launch()` can block for seconds if old thread mid-sleep.

**Worse:** If old thread stuck in slow `sync_iov()`:
```cpp
T0: Probe in middle of 30s disk I/O
T1: exit idle → stop()
T2: enter idle → launch()
      ↓ Blocks 30s waiting for I/O to complete!
```

**Possible mitigation:** Add timeout to probe I/O, or detach/don't wait.

---

## 🟢 Minor Issue #5: Probe Offset Not Updated After Swap

**Location:** `raid1.cpp:212-213`

**Problem:** `reserved_size` captured at RAID construction, not updated after `swap_device()`.

```cpp
_idle_probe_a.launch(state.active_dev, reserved_size);  // ← Old reserved_size
```

If swapped device has different size, probes read wrong offset.

**Impact:** Minor - probes test wrong location but likely still detect failures.

---

## Questions for Author

1. **Multi-queue support:** Does ublkpp support `nr_hw_queues > 1`? If yes, Issue #3 needs addressing.

2. **Thread safety assumptions:** Is `idle_transition()` guaranteed to be called from a single thread, or can multiple queues call it concurrently?

3. **Launch/stop lifecycle:** Is there an expectation that `launch()` should be fast/non-blocking? Current design can block for seconds.

4. **Swap semantics:** Should `swap_device()` wait for idle probes to fully stop before returning? Current implementation just requests stop.

---

## Testing Recommendations

1. **Stress test:** Rapid idle enter/exit cycles while I/O flowing
2. **Multi-queue:** Test with `nr_hw_queues=4`, concurrent idle transitions
3. **Swap during idle:** Swap device while probes running
4. **TSan validation:** Run with ThreadSanitizer to catch races (#1 and #2 should show up)

---

**Note:** These are potential issues identified through static analysis. Some may not manifest depending on actual ublksrv threading model and guarantees. Recommend reviewing with knowledge of ublksrv's queue threading behavior.

---

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

Ilyin-Ilya:
## Update: Race Condition Analysis - Findings

After comprehensive codebase analysis, I have critical findings about the race conditions I reported:

---

## 🔴 ACTUAL CRITICAL ISSUE: idle_transition() Not Wired Up!

**Discovery:** PR #184 adds `idle_transition()` override to `Raid1DiskImpl`, but **the target layer never calls it**.

### Current State (Before & After PR)

**Target layer** (`ublkpp_tgt.cpp:440,471`):
```cpp
static void idle_transition(ublksrv_queue const*, bool enter) { 
    TLOGT("Idle Trans: {}", enter)  // ← Only logs, never calls disk!
}

// Registered as callback:
.idle_fn = idle_transition,  // Called by ublksrv when I/O stops
```

**This static function:**
- ✅ Gets called by ublksrv library when queue goes idle
- ❌ **Does NOT call `disk->idle_transition(q, enter)`**
- ❌ Idle probes **never launch** in production

### What This Means

The entire idle probe mechanism in this PR is **unreachable**:

```cpp
// raid1.cpp:196 - This function is NEVER called!
void Raid1DiskImpl::idle_transition(ublksrv_queue const*, bool enter) {
    // ... immediate probe ...
    _idle_probe_a.launch(state.active_dev, reserved_size);
    _idle_probe_b.launch(state.backup_dev, reserved_size);
}
```

**Result:** UNAVAIL devices stay marked UNAVAIL forever (until manual recovery or service restart).

### How to Fix

**Option 1: Wire it up in target layer**

```cpp
// ublkpp_tgt.cpp
static void idle_transition(ublksrv_queue const* q, bool enter) { 
    TLOGT("Idle Trans: {}", enter)
    
    auto tgt = static_cast<ublkpp_tgt_impl*>(q->dev->tgt.tgt_data);
    if (tgt && tgt->_disk) {
        tgt->_disk->idle_transition(q, enter);  // ← Add this!
    }
}
```

**Option 2: Use existing tests pattern**

The tests manually call `raid_device.idle_transition(nullptr, true/false)`. This works but bypasses the ublksrv integration.

---

## Threading Model Findings

Based on codebase analysis:

### Default Configuration (nr_hw_queues=1)
- ✅ **Single queue thread** by default
- ✅ **No concurrent access** to UblkDisk methods
- ✅ **Race conditions #1-#4 cannot occur** with default config

### Multi-Queue Configuration (nr_hw_queues > 1)
- ⚠️ **One OS thread per queue** (one-to-one mapping)
- ⚠️ **All threads share same UblkDisk object**
- ⚠️ **No locks protect UblkDisk methods** (lock-free design)
- 🔴 **All reported races ARE REAL** if nr_hw_queues > 1

**From `ublkpp_tgt.cpp:78-108`:**
```cpp
sisl::named_thread(fmt::format("q_{}_{}", dev_id, i), [q] {
    ublksrv_queue_init_flags(q, ..., IORING_SETUP_SINGLE_ISSUER);
    while (!ublksrv_queue_is_done(q)) {
        ublksrv_process_io(q);  // Each queue thread processes I/O
    }
});
```

Each queue thread can call:
- `disk->async_iov()` (concurrent)
- `disk->on_io_complete()` (concurrent)
- `disk->idle_transition()` (concurrent, if wired up)

### Existing Thread Safety

**What IS protected:**
- `_read_route_cache`: atomic CAS
- Bitmap operations: atomic page allocation
- `__swap_device`: `_swap_lock` mutex

**What is NOT protected:**
- `_pending_results`: plain std::map (relies on queue-keyed access)
- `_resync_task`: shared pointer (no atomic updates)
- `_resync_enabled`: plain bool (known issue #180)
- **`idle_transition()` internals**: no synchronization

---

## Verdict on Reported Issues

| Issue | Real with nr_hw_queues=1? | Real with nr_hw_queues>1? | Actual Impact |
|-------|---------------------------|---------------------------|---------------|
| **#1: Concurrent idle_transition** | ❌ No | ✅ **YES** | **BLOCKED by wiring bug** |
| **#2: launch() not thread-safe** | ❌ No | ✅ **YES** | **BLOCKED by wiring bug** |
| **#3: Multi-queue idle tracking** | N/A | ✅ **YES** | **BLOCKED by wiring bug** |
| **#4: launch() blocks on join** | ✅ **YES** | ✅ **YES** | **BLOCKED by wiring bug** |
| **#5: Probe offset after swap** | ✅ **YES** | ✅ **YES** | **BLOCKED by wiring bug** |

**Key finding:** All issues are blocked by the fact that `idle_transition()` is never called from production code.

---

## Recommendations

### Immediate (Fix This PR)

1. **Wire up idle_transition in target layer:**
   ```cpp
   static void idle_transition(ublksrv_queue const* q, bool enter) {
       TLOGT("Idle Trans: {}", enter)
       auto tgt = static_cast<ublkpp_tgt_impl*>(q->dev->tgt.tgt_data);
       if (tgt && tgt->_disk) {
           tgt->_disk->idle_transition(q, enter);
       }
   }
   ```

2. **Add integration test** that verifies idle probes actually launch (not just unit test).

### Multi-Queue Safety (If nr_hw_queues > 1 is Supported)

If multi-queue is intended to work:

1. **Add `_idle_probes_active` atomic flag** (Issue #1)
2. **Add `_probe_mutex` to Raid1AvailProbeTask** (Issue #2)
3. **Track per-queue idle count** (Issue #3)
4. **Document that nr_hw_queues=1 is required** for RAID1

### Test Coverage Gaps

**Missing tests:**
- Integration test verifying ublksrv `idle_fn` → `UblkDisk::idle_transition()` chain
- Concurrent `idle_transition()` calls (if multi-queue)
- Idle probe actually launching in response to I/O stop
- Recovery detection via idle probe (not manual call)

---

## Current PR Status

**Functionality:** ⚠️ **Incomplete** - idle probes never launch in production

**Thread Safety:** 
- ✅ Safe with default nr_hw_queues=1
- ❌ Unsafe with nr_hw_queues > 1 (if wired up)

**Suggested Action:**
1. Fix wiring issue (add disk call in target layer)
2. Add integration test
3. Document nr_hw_queues=1 requirement (or add multi-queue safety)
4. Re-test with wired-up implementation

---

**Apologies for the initial incomplete analysis.** The race conditions I identified are real for multi-queue configurations, but the primary issue is that the feature is **not wired up** and therefore **not functional** as written.

---

🤖 Updated analysis by [Claude Code](https://claude.com/claude-code)

Ilyin-Ilya:
if nr_queues = 1, I believe we're good

Ilyin-Ilya:
## CORRECTION: Wiring Issue - My Mistake!

I apologize - I was **completely wrong** about the wiring issue. 

The PR **DOES** properly wire up `idle_transition()` in the target layer:

```cpp
// src/target/ublkpp_tgt.cpp (lines 1078-1082 in diff)
static void idle_transition(ublksrv_queue const* q, bool enter) {
    TLOGT("Idle Trans: {}", enter)
    auto device = reinterpret_cast<UblkDisk*>(q->dev->tgt.tgt_data);
    device->idle_transition(q, enter);  // ← Properly wired!
}
```

**This means:**
- ✅ Feature IS functional
- ✅ Idle probes WILL launch in production
- ⚠️ Race conditions I identified ARE relevant

---

## Updated Assessment

### Threading Issues Are Real (If nr_hw_queues > 1)

With the wiring in place, all the concurrency issues I identified **are actual concerns** if multi-queue is enabled:

| Issue | Severity | Real? |
|-------|----------|-------|
| #1: Concurrent idle_transition() race | 🔴 Critical | ✅ YES (if multi-queue) |
| #2: launch() not thread-safe | 🔴 Critical | ✅ YES (if multi-queue) |
| #3: Multi-queue idle tracking | 🟡 Important | ✅ YES (if multi-queue) |
| #4: launch() blocks on join | 🟡 Important | ✅ YES (always) |
| #5: Wrong probe offset after swap | 🟢 Minor | ✅ YES (always) |

### Default Configuration (nr_hw_queues=1)

With the default single-queue configuration:
- ✅ **All issues avoided** (single thread, no races)
- ✅ Feature works correctly
- ✅ Safe to merge as-is for default config

### Multi-Queue Configuration (nr_hw_queues > 1)

If ublkpp supports multi-queue:
- 🔴 Issues #1 and #2 are **data races** (UB)
- 🟡 Issues #3-5 are **logic bugs**
- ⚠️ Need synchronization fixes

---

## Recommendations (Updated)

### If nr_hw_queues=1 Only (Current Default)
- ✅ **Approve as-is** - feature works, no races possible
- 📝 Document that RAID1 requires nr_hw_queues=1

### If Multi-Queue Support Intended
- 🔧 Add synchronization (atomic flag + mutex)
- 🧪 Add concurrent stress tests
- 📝 Document thread-safety guarantees

---

**My sincere apologies for the confusion.** I was analyzing the current `main` branch state instead of the PR diff changes. The PR is properly implemented - my concerns about races only apply if multi-queue configurations are used.

Given the default nr_hw_queues=1, the PR is **safe and functional as written**.

---

🤖 Corrected by [Claude Code](https://claude.com/claude-code) - embarrassed but accurate now

szmyd:
Thanks for the thorough follow-up analysis. Addressing each point:

---

**#1, #2, #3 — Concurrent `idle_transition()`, `launch()` thread safety, multi-queue idle tracking**

You correctly self-corrected these: with `nr_hw_queues=1` (the default, and the only supported configuration for RAID1) there is exactly one queue thread, so none of these concurrent-access scenarios are possible. All `UblkDisk` methods are called from the single queue thread; this is a deliberate design constraint, not an oversight. If multi-queue support is added in the future the idle probe would need revisiting, but that's out of scope here.

---

**#4 — `launch()` blocking on join**

The sleep path is not a concern: line 36 uses the stop-token-aware overload of `condition_variable_any::wait_for`:

```cpp
cv.wait_for(lk, st, delay, [] { return false; });
```

When the old `jthread`'s destructor calls `request_stop()`, the stop token fires its registered callback which immediately notifies the CV — the sleep wakes at once, not after the full `avail_delay`. The only scenario where `launch()` could block is if the probe thread is inside `sync_iov()` at the moment stop is requested. In practice this is fine: `stop()` is called on idle-exit (I/O arriving), and `launch()` is not called again until the next idle-enter, which requires all in-flight user I/O to drain first. Any in-progress `sync_iov` will have long completed by then.

---

**#5 — `reserved_size` after swap**

`reserved_size` is a member of `Raid1DiskImpl`, not a construction-time capture. It reflects the current array geometry. More importantly, `swap_device()` already stops the probes; the next `idle_transition(enter=true)` relaunches them with the current member value. No stale offset is possible.

---

**Wiring issue**

You corrected this yourself — `ublkpp_tgt.cpp:442-443` does call `device->idle_transition(q, enter)`. The feature is fully wired.

---

No code changes needed from this round. The existing implementation handles all these cases correctly within its single-queue constraint.

Ilyin-Ilya (commented):
did not address about the non-used code. 
Did not address comment about info logging.
No changelog.md changes

szmyd:
The multi-queue concurrency concerns raised during review (concurrent `idle_transition()`, `launch()` thread safety, idle tracking across queues) are tracked as future work in #193. The implementation here is correct and safe under the `nr_hw_queues=1` constraint.

src/raid/raid1/raid1.cpp
szmyd: Won't change, we never really look at logs unless we're "debugging" an issue. This is made visible to the OM (the whole point), if we want a bouncing audit, OM is the place to check.

szmyd:
> did not address about the non-used code.

yes it was removed.

> Did not address comment about info logging.

Yes, comment made and resolved.

> No changelog.md changes

Valid, will do

Ilyin-Ilya (approved):

PR #194, opened 2026-04-13
raid1: Fix defunct device remount failure

szmyd:
## Problem

A degraded RAID-1 volume mounted via `forcedPartialExpose` (one iSCSI endpoint missing → `DefunctDisk`) works on first attach but fails fatally on every subsequent remount:

```
[error] [raid1_superblock.cpp:69] Error writing Superblock to: [~DEFUNCT~, size=17179869183Gi, lbs=0x200]
[error] [tgt_mgr.cpp:218:add_raid10] Failed to construct RAID-10 Array: Could not initialize superblocks!
```

## Root Cause

`__init_bitmap_and_degraded_route()` used `state.active_dev->new_device` (from `__capture_route_state()`) to decide whether to route reads to `DEVA` or `DEVB`. The problem is that `active_dev` is *role-relative* (whichever device is the current read target), not *slot-relative*, while `read_route::DEVA/DEVB` refer to physical slots.

**First mount** (defunct is `_device_a`, persisted route = `EITHER`):
- `__capture_route_state()` → `active_dev = _device_a` (defunct, `new_device=true`)
- `true ? DEVB : DEVA` → `DEVB` ✓ — routes to the live `_device_b`

**Remount** (defunct is still `_device_a`, persisted route = `DEVB`):
- `__capture_route_state()` with `route=DEVB` → `active_dev = _device_b` (live), `backup_dev = _device_a` (defunct)
- `DEFUNCT_DEVICE(backup_dev)` → true, enters defunct branch
- `active_dev->new_device = false` (live device has a valid superblock)
- `false ? DEVB : DEVA` → route overwritten to **`DEVA`** ← points to `DefunctDisk`
- `__become_active()` tries to `write_superblock(*defunct_disk)` → `io_error` → **throws**

The same inversion occurs symmetrically when `_device_b` is the defunct slot.

## Fix

Replace the role-relative `__capture_route_state()` lookup with direct references to `_device_a` and `_device_b` physical slots. Since `read_route::DEVA/DEVB` map directly to physical slots, the defunct check and route selection are now unambiguous regardless of which route was persisted in the superblock:

```cpp
bool const a_is_defunct = DEFUNCT_DEVICE(_device_a->disk);
__store_read_route(a_is_defunct ? read_route::DEVB : read_route::DEVA);
_dirty_bitmap->load_from(*(a_is_defunct ? _device_b : _device_a)->disk);
```

The same physical-slot approach is applied to the last `else if` (degraded + clean remount) branch, which also used `state.active_dev/backup_dev`.

## Tests

Two regression tests added to `superblock/defunct_disk.cpp`:

- **`DefunctRemountA`** — `device_a` is defunct, `device_b` returns a persisted superblock with `read_route=DEVB`. Asserts no throw and that `__become_active` writes to `device_b` (not `device_a`) with the route preserved.
- **`DefunctRemountB`** — symmetric case: `device_b` is defunct, persisted `read_route=DEVA`, asserts write goes to `device_a`.

All 140 Raid1 unit tests pass.

### DIFF
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -214,8 +214,7 @@ void Raid1DiskImpl::__init_bitmap_and_degraded_route() {
         _sb->fields.bitmap.age = htobe64(be64toh(_sb->fields.bitmap.age) + 16);
         RLOGW("Unclean shutdown in degraded mode! Dirty all of BITMAP")
         _dirty_bitmap->dirty_region(0, capacity());
-    } else if (read_route::EITHER != __get_read_route()) {
-        auto const route = __get_read_route();
+    } else if (auto const route = __get_read_route(); read_route::EITHER != route) {
         auto const& active_dev = (route == read_route::DEVB) ? _device_b : _device_a;
         auto const& backup_dev = (route == read_route::DEVB) ? _device_a : _device_b;
         RLOGW("Raid1 is starting in degraded mode [uuid:{}]! Degraded device: {}", _str_uuid, *backup_dev->disk)
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -184,37 +184,42 @@ void Raid1DiskImpl::__load_and_select_superblock(boost::uuids::uuid const& uuid,
 }
 
 void Raid1DiskImpl::__init_bitmap_and_degraded_route() {
-    auto const state = __capture_route_state();
     // Read in existing dirty BITMAP pages
     _dirty_bitmap = std::make_shared< Bitmap >(capacity(), be32toh(_sb->fields.bitmap.chunk_size), block_size(),
                                                _sb->superbitmap_reserved, _str_uuid);
-    if (state.active_dev->new_device) _dirty_bitmap->init_to(state.active_dev->disk);
-    if (state.backup_dev->new_device) _dirty_bitmap->init_to(state.backup_dev->disk);
+    // Initialize bitmap pages for any new (or defunct) device slots
+    if (_device_a->new_device) _dirty_bitmap->init_to(_device_a->disk);
+    if (_device_b->new_device) _dirty_bitmap->init_to(_device_b->disk);
 
-    // We need to completely dirty one side if either is new when the other is not, this will also
-    // apply to when a DefunctDisk is provided, load_from will be skipped in this case.
-    if ((DEFUNCT_DEVICE(state.active_dev->disk) || DEFUNCT_DEVICE(state.backup_dev->disk))) {
+    // Use physical slot references (_device_a/_device_b) directly to avoid the ambiguity of
+    // role-relative state captured by __capture_route_state(). The read_route enum refers to
+    // physical slots (DEVA=_device_a, DEVB=_device_b), so mapping must be slot-based.
+    if (DEFUNCT_DEVICE(_device_a->disk) || DEFUNCT_DEVICE(_device_b->disk)) {
         RLOGW("RAID1 device [uuid:{}] is running with a defunct device!", _str_uuid)
-        // Defunct disks are marked NEW
-        __store_read_route(state.active_dev->new_device ? read_route::DEVB : read_route::DEVA);
-        // We still want to load from the BITMAP and not bump the age, maybe we'll get the orig
-        // disk back in a swap when it becomes available?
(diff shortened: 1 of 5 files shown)


Ilyin-Ilya:
## Code Review

### Assessment
- 🔴 Critical Issues: 0
- 🟡 Important Issues: 1
- 🟢 Minor Issues: 2

### Strengths

- Excellent root cause analysis in the PR description — the role-relative vs. physical-slot mismatch is clearly explained with the two concrete mount sequences
- Removing the `__capture_route_state()` abstraction from `__init_bitmap_and_degraded_route()` is the right architectural call; the abstraction was hiding the exact semantic mismatch that caused the bug
- Good inline comments explaining *why* physical slot references are used — future readers will understand the invariant
- Regression tests verify the key invariant end-to-end: the SB write lands on the live device and preserves the `read_route` value

---

### Issues Found

#### 🟡 Important

**CHANGELOG description only mentions slot A (`src/CHANGELOG.md`)**

> "Fix the bringing up a degraded array with a defunct disk in slot A."

The fix actually covers both slot A and slot B symmetrically (the `bool const a_is_defunct` logic and the `DefunctRemountB` test confirm this). The wording implies only one direction was broken, which is misleading and makes it harder to trace the fix later. Suggested wording:

> "raid1: Fix remount failure when bringing up a degraded array with a defunct device"

---

#### 🟢 Minor

**Double call to `__get_read_route()` in the degraded-clean-remount branch (`src/raid/raid1/raid1.cpp`)**

```cpp
} else if (read_route::EITHER != __get_read_route()) {
    auto const route = __get_read_route();   // ← called a second time
    auto const& active_dev = (route == read_route::DEVB) ? _device_b : _device_a;
```

Not a correctness issue in single-threaded init, but the value is already computed for the condition. An if-with-initializer eliminates the redundancy:

```cpp
} else if (auto const route = __get_read_route(); read_route::EITHER != route) {
    auto const& active_dev = (route == read_route::DEVB) ? _device_b : _device_a;
```

---

**`EXPECT_TO_WRITE_SB` placement in regression tests is non-obvious (`src/raid/raid1/tests/superblock/defunct_disk.cpp`)**

In `DefunctRemountA` and `DefunctRemountB`, the explicit `EXPECT_CALL` on `UBLK_IO_OP_WRITE` covers the `__become_active()` write during construction, and then `EXPECT_TO_WRITE_SB(device_b/a)` is called *after* construction to cover the destructor's clean-unmount write. This is functionally correct (I checked `EXPECT_TO_WRITE_SB` expands to a `EXPECT_CALL(...).Times(1)`, and it fires before `raid_device` leaves scope), but the ordering looks like a post-construction assertion to readers unfamiliar with this pattern.

A one-line comment would help:

```cpp
auto raid_device = ublkpp::Raid1Disk(...);
// Set up destructor's clean-unmount write expectation (fires before raid_device leaves scope)
EXPECT_TO_WRITE_SB(device_b);
```

---

### Recommendations

None blocking. The fix is correct and well-reasoned. The CHANGELOG wording is the only thing worth updating before merge.

---
🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

_Note: This is a review comment only. Approval/rejection will be done manually._

Ilyin-Ilya (dismissed):

src/raid/raid1/raid1.cpp
```diff
@@ -184,37 +184,42 @@ void Raid1DiskImpl::__load_and_select_superblock(boost::uuids::uuid const& uuid,
...
     _dirty_bitmap = std::make_shared< Bitmap >(capacity(), be32toh(_sb->fields.bitmap.chunk_size), block_size(),
                                                _sb->superbitmap_reserved, _str_uuid);
-    if (state.active_dev->new_device) _dirty_bitmap->init_to(state.active_dev->disk);
-    if (state.backup_dev->new_device) _dirty_bitmap->init_to(state.backup_dev->disk);
-
-    // We need to completely dirty one side if either is new when the other is not, this will also
-    // apply to when a DefunctDisk is provided, load_from will be skipped in this case.
-    if ((DEFUNCT_DEVICE(state.active_dev->disk) || DEFUNCT_DEVICE(state.backup_dev->disk))) {
+    // Initialize bitmap pages for any new (or defunct) device slots
+    if (_device_a->new_device) _dirty_bitmap->init_to(_device_a->disk);
```
Ilyin-Ilya, line 191: note: called in constructor, so are safe to access

Ilyin-Ilya (approved):

PR #198, opened 2026-04-16
raid1: Fix resync task hang and launch race conditions

szmyd:
## Summary

- **Stop hang during unavail wait**: `stop()` returned `SUCCESS` for `IDLE` state without CASing to `STOPPING`, so `join()` blocked forever when the resync thread was spinning in the unavail loop. Fixed by retrying `IDLE→STOPPING` CAS in `stop()` and adding an explicit `__load_state()` check in the unavail branch of `_start()`.
- **Concurrent launch race**: `launch()` used `CAS(IDLE→IDLE)` which provides no mutual exclusion — two concurrent callers could both pass and race on `_resync_task = std::thread{...}`, triggering `std::terminate()`. Fixed with `_launch_lock` mutex in both `launch()` and `stop()`.
- **Incorrect degraded device logged**: `__become_degraded` always logged `backup_dev` in the `RLOGW`; now logs `failed_device` correctly for both degradation directions.

Includes regression tests for both concurrency bugs (`StopDuringUnavailWait`, `ConcurrentLaunch`) using `Raid1ResyncTask` directly, following the pattern established in `resync_relaunch_after_complete.cpp`.

## Test plan

- [x] `Raid1Concurrency.StopDuringUnavailWait` — marks mirror_b unavail with always-failing probe reads, calls `stop()` from a separate thread, asserts completion within 15s
- [x] `Raid1Concurrency.ConcurrentLaunch` — fires two simultaneous `launch()` calls via `std::barrier`, asserts no crash
- [x] All existing RAID1 concurrency tests continue to pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.21.4
+- raid1: Fix resync task hang when stop() is called during unavail wait loop
+- raid1: Fix concurrent launch()/stop() race on resync task thread assignment
+- raid1: Fix incorrect device logged as degraded in __become_degraded (DEVB route)
+
 ## 0.21.3
 - raid1: Fix Raid1ResyncTask::launch() crash on joinable thread reassignment
 
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=2.0"
 
 class UBlkPPConan(ConanFile):
     name = "ublkpp"
-    version = "0.21.3"
+    version = "0.21.4"
 
     homepage = "https://github.com/szmyd/ublkpp"
     description = "A UBlk library for CPP application"
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -604,10 +604,12 @@ io_result Raid1DiskImpl::__become_degraded(sub_cmd_t failed_path, RouteState con
     }
 
     auto const backup_clean = (read_route::DEVB == new_route);
+    auto& failed_device = backup_clean ? cur_state->active_dev : cur_state->backup_dev;
+    auto& working_device = backup_clean ? *cur_state->backup_dev->disk : *cur_state->active_dev->disk;
 
     auto const old_age = _sb->fields.bitmap.age;
     _sb->fields.bitmap.age = htobe64(be64toh(_sb->fields.bitmap.age) + 1);
-    RLOGW("Device became degraded {} [age:{}] [uuid:{}]", *cur_state->backup_dev->disk,
+    RLOGW("Device became degraded {} [age:{}] [uuid:{}]", *failed_device->disk,
           static_cast< uint64_t >(be64toh(_sb->fields.bitmap.age)), _str_uuid);
 
     // Record degradation event in metrics with device name
@@ -617,7 +619,6 @@ io_result Raid1DiskImpl::__become_degraded(sub_cmd_t failed_path, RouteState con
     }
 
     // Must update age first; we do this synchronously to gate pending retry results
-    auto& working_device = backup_clean ? *cur_state->backup_dev->disk : *cur_state->active_dev->disk;
     if (auto sync_res = write_superblock(working_device, _sb.get(), backup_clean, new_route); !sync_res) {
         // Rollback the failure to update the header
         _sb->fields.bitmap.age = old_age;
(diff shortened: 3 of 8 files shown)


d2chau:
Review feedback from claude
```
  This PR fixes three critical RAID1 resync task bugs that could cause system hangs and crashes in production. The changes address concurrency issues in resync task lifecycle
  management and fix incorrect logging during device degradation.

  Version: Bumps from 0.21.3 → 0.21.4
  Impact: High - fixes potential deadlocks and race conditions in RAID1 recovery
  Test Coverage: +0.44% overall (59.91% → 60.36%), with two new regression tests

  ---
  Critical Fixes Analysis

  1. ✅ Resync Task Hang Fix (stop() deadlock)

  Problem: When stop() was called while _start() was spinning in the unavailable device loop, the state machine could get stuck:
  - Thread spinning with state=IDLE waiting for device
  - stop() handler returned early for IDLE without transitioning to STOPPING
  - join() blocked forever waiting for thread that never checked stop condition

  Solution: Restructured _start() to check STOPPING state inside the unavail wait loop and added RETRY_WITH_SLEEP fallback in stop() for IDLE state with joinable thread.

  Quality: ✅ Excellent
  - Fixes root cause rather than papering over symptom
  - Unified wait loop is cleaner than previous two-stage approach
  - Test validates fix (StopDuringUnavailWait ensures completion within 15s)

  Concern: ⚠️ The fallthrough pattern in stop() is subtle:
  case resync_state::IDLE: {
      if (_resync_task.joinable()) return {state, transition_action::RETRY_WITH_SLEEP};
      [[fallthrough]];
  }
  This assumes non-joinable IDLE means no thread to stop. Consider adding a comment explaining this invariant.

  ---
  2. ✅ Concurrent Launch Race Fix

  Problem: Multiple threads calling launch() simultaneously could both pass the CAS(IDLE→IDLE) check (which provides zero mutual exclusion) and race on _resync_task assignment,
  causing std::terminate() when destroying a running thread.

  Solution: Added _launch_lock mutex to serialize launch() and stop().

  Quality: ✅ Correct but conservative
  - Simple and foolproof solution
  - stop() also takes lock to prevent launch/stop interleaving

  Minor Concern: Lock is held during entire launch() including thread creation. If thread startup is slow, this could delay concurrent operations. However, given that concurrent
  launches are likely rare and should be serialized anyway, this is acceptable.

  Test Coverage: ✅ ConcurrentLaunch test validates via std::barrier synchronization.

  ---
  3. ✅ Degraded Device Logging Fix

  Problem: __become_degraded() always logged backup_dev as failed, but when backup_clean=true (DEVB route), the active_dev was actually the degraded one.

  Solution: Moved variable declarations before log statement and selected correct device based on backup_clean:
  auto& failed_device = backup_clean ? cur_state->active_dev : cur_state->backup_dev;

  Quality: ✅ Straightforward correctness fix
  - Pure logging bug with no functional impact
  - Makes troubleshooting much easier

  ---
  Code Quality Assessment

  Strengths

  - Clear intent: Each fix targets a specific, well-documented bug
  - Defensive programming: Added regression tests to prevent future breaks
  - Minimal scope: Changes are surgical - no unnecessary refactoring
  - Improved structure: Unified wait loop in _start() is more maintainable

  Style & Conventions

  ✅ Follows ublkpp patterns:
  - RLOG macros for logging
  - Atomic state machine with CAS operations
  - SISL_OPTIONS for configuration

  Potential Issues

  1. Lock Scope in stop() - The _launch_lock is held while potentially sleeping in RETRY_WITH_SLEEP. Consider if this could starve other operations:
  uint32_t Raid1ResyncTask::stop() noexcept {
      auto lg = std::scoped_lock< std::mutex >(_launch_lock);  // Held during retries
      __transition_to(resync_state::PAUSE, resync_state::STOPPING, [this](resync_state state) {
          // May sleep and retry multiple times with lock held
      });
  }
  Recommendation: Document that stop() is expected to be rare and brief, or consider releasing lock during sleep.

  2. Test Coverage Gap - The diff shows 45.45% patch coverage (18 lines missing). What paths aren't tested?
    - Are the edge cases in the unified wait loop covered?
    - What about the fallthrough path in stop()?
  3. Memory Ordering - The _start() method reads unavail with acquire ordering but loads state with relaxed (via __load_state()). Is this intentional?
  while (dirty_mirror->unavail.test(std::memory_order_acquire) || ...) {
      cur_state = __load_state();  // What memory order is this?
  Recommendation: Verify __load_state() uses appropriate memory order for inter-thread visibility.

  ---
  Performance Implications

  Positive:
  - Unified wait loop reduces redundant state checks
  - No new allocations or performance regressions

  Neutral:
  - Mutex adds minimal overhead since launch/stop are infrequent operations
  - Lock contention unlikely in practice (resync happens once per degradation event)

  ---
  Security Considerations

  ✅ No security implications - internal state management only.

  ---
  Testing Recommendations

  Covered ✅

  - Concurrent launch race
  - Stop during unavail wait

  Missing Gaps to Consider

  1. Stop during active resync - Does stop work correctly when resync is ACTIVE and making progress?
  2. Repeated launch after stop - Can we launch → stop → launch cleanly?
  3. Device becomes available during stop - What if device comes back online while stopping?

  Consider expanding regression suite to cover these scenarios.

  ---
  Final Recommendation

  ✅ APPROVE with minor suggestions

  This PR fixes legitimate, high-severity bugs with appropriate testing. The code is clear and follows project conventions. Suggested improvements:

  1. Add comment explaining the IDLE fallthrough invariant in stop()
  2. Verify memory ordering in unified wait loop is correct
  3. Document lock scope in stop() or optimize to release during sleep
  4. Investigate 18 uncovered lines - are they dead code or missing test cases?
  ```

d2chau (approved):
LGTM, maybe we should add some future enhancements based on the feedback

PR #200, opened 2026-04-20
We logged every RAID1 shutdown indicating the resync tasked was stopped.

szmyd:
This is confusing when the array is CLEAN...only log if it is actually running and is stopped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.21.5
+- raid1: Only log in reference to Resync if it was actually running.
+
 ## 0.21.4
 - raid1: Fix resync task hang when stop() is called during unavail wait loop
 - raid1: Fix concurrent launch()/stop() race on resync task thread assignment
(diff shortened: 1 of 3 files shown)


szmyd:
## Code Review: Commit 6192746

### Overview
Removes a misleading `RLOGI("Terminating Resync Task")` from `stop()` that fired unconditionally on every RAID1 shutdown — even when the array was clean and no resync was ever running. Simultaneously promotes the "Resync Task Stopped" log in `_start()` from `RLOGD` → `RLOGI`, so the event is still visible at INFO level when resync **was** actually in-flight.

---

### Analysis

**Correctness** — Solid.
- `stop()` is always called on shutdown regardless of resync state. Removing the log there eliminates the false positive on clean arrays.
- The `RLOGD` → `RLOGI` promotion in `_start()` (line 112) is correct: that branch is only reached when the resync thread was actually executing and detected `STOPPING`. This guarantees INFO visibility for real stops with zero false positives.
- No regression in observability: mid-resync stops still produce the INFO log.

**One gap worth noting:** `stop()` now logs nothing at any level. If resync IS running and gets stopped, the only trace is from `_start()` — which is correct, but means there's no log from `stop()` itself even at DEBUG. Not a bug, but something to be aware of if future debugging requires tracing the full shutdown sequence.

**Style / Conventions** — Follows project conventions.

**Tests** — None needed; logging-only change.

**Version / Changelog** — Correctly bumped patch version (0.21.4 → 0.21.5), CHANGELOG entry present.

---

### Verdict

Clean, minimal, and correct. No issues.

Ilyin-Ilya (approved):

PR #201, opened 2026-04-21
raid1: Fix multi-queue support (#193)

### LINKED ISSUE #193: raid1: Support multi-queue (nr_hw_queues > 1) for idle probe and UNAVAIL state
## Background

PR #184 introduced the UNAVAIL replica state and an idle-scoped periodic health probe (`Raid1AvailProbeTask`). The implementation is deliberately scoped to the current default of `nr_hw_queues=1`, where all `UblkDisk` methods are called from a single queue thread and no concurrent-access issues can arise.

If RAID1 is ever extended to support multi-queue (`nr_hw_queues > 1`), the idle probe requires significant rework.

### Critical (data races / UB with multi-queue)

1. **Concurrent `idle_transition()` calls** — Multiple queue threads would call `idle_transition()` concurrently with no synchronization. `_idle_probe_a/b.stop()` and `.launch()` are not thread-safe across callers.

2. **`launch()` not thread-safe** — `Raid1AvailProbeTask::launch()` starts a `jthread` and calls `.join()` on any prior thread without a lock. Concurrent launches would race on the `jthread` member.

### Logic bugs with multi-queue

3. **Multi-queue idle tracking** — The idle probe should only run when *all* queues are idle. Currently `idle_transition(enter=true)` from any single queue would start the probe, even while other queues are still active. Correct behavior requires a counter or bitmask of idle queues, starting the probe only when the count reaches `nr_hw_queues`.

## Proposed Fix

- Add an atomic idle-queue counter; start the probe when all queues report idle, stop on first active-queue transition.
- Guard `Raid1AvailProbeTask::launch()` / `stop()` with a mutex (or make the probe start/stop idempotent via an atomic flag).
- Add concurrent stress tests exercising the multi-queue idle path.
- Document that `nr_hw_queues=1` is the current supported configuration for RAID1, and update that documentation when this issue is resolved.

## Out of scope for PR #184

This work is explicitly out of scope for #184. The single-queue implementation in that PR is correct and safe. This issue tracks future work if multi-queue RAID1 support is desired.

szmyd:
Fix three classes of races exposed when running with nr_hw_queues > 1.

## RAID-1: multi-queue idle probe (issue #193)

The idle probe was written assuming nr_hw_queues=1. With multiple queues:

- Any single queue going idle started the probe; any single queue exiting idle
  stopped it — wrong semantics under N queues.
- Concurrent idle_transition() calls from different queue threads raced on the
  jthread members of Raid1AvailProbeTask (unguarded move-assignment).
- swap_device() called stop() on the probes without holding _idle_probe_lock,
  racing against idle_transition()'s launch().
- idle_transition() called __capture_route_state() outside _idle_probe_lock,
  so swap_device() could replace the device between the capture and launch(),
  leaving a background probe running against the outgoing (now-swapped) device.

Fix: add _idle_queue_count (atomic) that gates probe start until all N queues
report idle and stops on the first exit. _idle_probe_lock (mutex) serializes all
launch()/stop() calls. _nr_hw_queues is incremented by open_for_uring() (called
once per queue thread from init_queue), so the threshold is set before any I/O
or idle_transition() fires.

idle_transition() now uses two route-state captures: one pre-lock for an
immediate synchronous probe (fire-and-forget, clears UNAVAIL on already-recovered
devices without delay), and a second capture inside _idle_probe_lock for the
periodic launch() so it always uses the device that is actually current.

A DEBUG_ASSERT guards the exit path against ublksrv contract violations
(exit without matching enter), which would cause _idle_queue_count to wrap and
permanently suppress probe launches.

## RAID-1: _pending_results access safety

_pending_results[q] used operator[] on the I/O path, which inserts on miss and
is not thread-safe under concurrent reads. open_for_uring() pre-populates the
map under _ctrl_lock during init; the I/O path uses find() + DEBUG_ASSERT (const
tree traversal, lock-free). _swap_lock renamed to _ctrl_lock to reflect that it
guards both swap_device() callers and _pending_results insertions.

## RAID-1: SuperBitmap::clear_all() races concurrent set_bit()

memset() may issue SIMD-wide stores that violate the per-byte atomicity contract
established by atomic_ref<uint8_t> in set_bit/clear_bit/test_bit. Write I/Os
reach dirty_region() -> set_bit() without holding _ctrl_lock, so they race with
init_to() -> clear_all() during device swap. Replace memset with byte-by-byte
atomic_ref<uint8_t>::store(relaxed) to eliminate the UB.

Also adds explanatory comments above __swap_device and __become_degraded
documenting the CAS-based mutual exclusion between them (a recurring review
question): both CAS _read_route_cache via __set_read_route(); exactly one wins,
the loser returns early — no additional lock is needed between them.

## RAID-1: init_queue error handling

init_queue now returns -1 (triggering ublksrv_dev_deinit() teardown) instead of
silently discarding any FDs returned by open_for_uring(). FDs returned here
would go unregistered with io_uring and cause fixed-file I/O to crash; treating
this as a fatal init failure catches any future disk type that exports FDs.

## Breaking API change

UblkDisk::open_for_uring signature changed from (int) to (ublksrv_queue const*, int).
Out-of-tree subclasses must update their override.

## Test fixes

- SuperBitmapTest.ConcurrentReadWhileWrite failed spuriously in Release CI builds:
  the optimized writer completed all 100k set_bit() calls before reader threads
  were scheduled, leaving read_count=0. Fixed with std::barrier.
- Three concurrency tests (ResyncRelaunchAfterComplete, ConcurrentLaunch,
  StopDuringUnavailWait) constructed MirrorDevice with a bare sync_iov mock that
  returned success without writing to the buffer. posix_memalign does not
  zero-initialize, so if freed memory happened to contain valid magic bytes with a
  wrong UUID, load_superblock would fail the UUID check and throw. Fixed with a
  sync_iov_zero_on_read() helper that zeros the buffer on READ calls. Extended
  to MultiQueueIdleConcurrent and MultiQueueIdleRapidToggle for the same reason.
- IdleTransitionExit and IdleExitSkipsProbe called idle_transition(false) without
  a prior idle_transition(true), violating the ublksrv contract asserted by the
  new DEBUG_ASSERT. Fixed by ensuring enter always precedes exit in tests.
  IdleTransitionExit renamed to IdleTransitionRoundTrip to accurately describe
  what it exercises.
- MockUblksrv constructor populated _queues[] fields after calling open_for_uring,
  so any disk implementation that reads ring_ptr or dev inside open_for_uring
  would see null values. Fixed by swapping the loop order.
- MultiQueueIdle* tests passed nullptr as the queue pointer to open_for_uring,
  bypassing the _pending_results.emplace() path introduced by this PR. Fixed by
  passing real stack-allocated ublksrv_queue pointers.
- Added SwapDeviceWhileIdleTransitioning: a TSAN regression guard that races
  swap_device() against concurrent idle_transition() calls to catch any future
  removal of the locked state re-capture in the probe enter path.

Closes #193

### DIFF
--- a/src/raid/raid0/raid0.cpp
+++ b/src/raid/raid0/raid0.cpp
@@ -90,10 +90,10 @@ std::shared_ptr< UblkDisk > Raid0Disk::get_device(uint32_t stripe_offset) const 
     return _stripe_array[stripe_offset]->disk;
 }
 
-std::list< int > Raid0Disk::open_for_uring(int const iouring_device_start) {
+std::list< int > Raid0Disk::open_for_uring(ublksrv_queue const* q, int const iouring_device_start) {
     auto fds = std::list< int >();
     for (auto& stripe : _stripe_array) {
-        fds.splice(fds.end(), stripe->disk->open_for_uring(iouring_device_start + fds.size()));
+        fds.splice(fds.end(), stripe->disk->open_for_uring(q, iouring_device_start + fds.size()));
     }
     return fds;
 }
--- a/src/raid/raid0/raid0.cpp
+++ b/src/raid/raid0/raid0.cpp
@@ -1,5 +1,6 @@
 #include "ublkpp/raid/raid0.hpp"
 
+#include <bit>
 #include <boost/uuid/uuid_io.hpp>
 #include <ublksrv.h>
 #include <ublksrv_utils.h>
@@ -32,6 +33,9 @@ load_superblock(UblkDisk& device, boost::uuids::uuid const& uuid, uint32_t& stri
 Raid0Disk::Raid0Disk(boost::uuids::uuid const& uuid, uint32_t const stripe_size_bytes,
                      std::vector< std::shared_ptr< UblkDisk > >&& disks) :
         UblkDisk(), _stripe_size(stripe_size_bytes), _stride_width(_stripe_size * disks.size()) {
+    if (disks.size() > _max_stripe_cnt)
+        throw std::invalid_argument(
+            fmt::format("Raid0Disk: too many disks ({}), max is {}", disks.size(), _max_stripe_cnt));
     // Discover overall Device parameters
     auto& our_params = *params();
     our_params.types |= UBLK_PARAM_TYPE_DISCARD;
@@ -175,66 +179,53 @@ io_result Raid0Disk::handle_discard(ublksrv_queue const* q, ublk_io_data const* 
 //  stripe boundaries and even wrap around several strides. This routine handles this calculation and calls
 //  the given routine `func` for each stripe that it has collected scatter (struct iovec) operations for.
 io_result Raid0Disk::__distribute(iovec* iovecs, uint64_t addr, auto&& func, bool retry, sub_cmd_t sub_cmd) const {
-    // Each thread has a 2-dimensional block of iovecs that we can split into.
-    thread_local auto sub_cmds =
-        std::array< std::tuple< uint64_t, uint32_t, std::array< iovec, 16 > >, _max_stripe_cnt >();
-    // Reset only the stripes we touch (cheaper than zeroing all _max_stripe_cnt on every call).
-    // A previous call that returned early on error leaves non-zero alive_cmds behind; stale iov_base
-    // pointers in those entries would corrupt the next I/O on this thread if not cleared.
-    static_assert(_max_stripe_cnt <= 64, "dirty_mask requires _max_stripe_cnt <= 64");
+    struct StripeAccum {
+        uint64_t io_addr{0};
+        uint32_t alive_cmds{0};
+        std::array< iovec, 16 > io_array{};
+    };
+    // Reset only touched stripes on exit; a failed call leaves non-zero alive_cmds that would corrupt the next I/O.
+    static_assert(_max_stripe_cnt == 64, "dirty_mask must be exactly uint64_t");
+    thread_local auto sub_cmds = std::array< StripeAccum, _max_stripe_cnt >();
     uint64_t dirty_mask{0};
     struct DirtyGuard {
         decltype(sub_cmds)& cmds;
         uint64_t& mask;
         ~DirtyGuard() noexcept {
             while (mask) {
-                std::get< 1 >(cmds[std::countr_zero(mask)]) = 0;
-                mask &= mask - 1;
+                cmds[std::countr_zero(mask)].alive_cmds = 0;
+                mask &= mask - 1; // clear lowest set bit
             }
         }
-    } _guard{sub_cmds, dirty_mask};
+    } guard{sub_cmds, dirty_mask};
 
-    // Special case for single device
     if (1 == _stripe_array.size()) return func(0, sub_cmd, iovecs, 1, addr);
 
     auto const route_mask = _max_stripe_cnt - 1;
-
     DEBUG_ASSERT_LE(iovecs->iov_len, UINT32_MAX) // LCOV_EXCL_LINE
-    auto const len = (uint32_t)iovecs->iov_len;
+    auto const len = static_cast< uint32_t >(iovecs->iov_len);
     uint32_t cnt{0};
     for (auto off = 0U; len > off;) {
         auto const [stripe_off, logical_off, sz] =
-            raid0::next_subcmd(_stride_width, _stripe_size, addr + off, (len - off));
-
-        // Ensure we advance here in case we _continue_ or anything later
-        auto buf_cursor = (uint8_t*)iovecs->iov_base + off;
+            raid0::next_subcmd(_stride_width, _stripe_size, addr + off, len - off);
+        auto buf_cursor = static_cast< uint8_t* >(iovecs->iov_base) + off;
         off += sz;
 
-        // Get the device
         auto const& device = _stripe_array[stripe_off]->disk;
-
-        // If this is a retry, we only want to re-issue the operation whose route matches the one passed in
-        if (retry) [[unlikely]] {
-            // Mask off to get "our" portion of the original route and see if the device that processed this
-            // operation matches the current RAID-0 sub-operation; if not then skip.
-            if (stripe_off != ((sub_cmd >> device->route_size()) & route_mask)) continue;
-        }
+        // On retry, re-issue only the stripe whose route bits match the original sub_cmd.
+        if (retry && stripe_off != ((sub_cmd >> device->route_size()) & route_mask)) [[unlikely]]
+            continue;
 
         dirty_mask |= 1ULL << stripe_off;
-        auto& [io_addr, alive_cmds, io_array] = sub_cmds[stripe_off];
-        { // Fillout iovec
-            auto& iov = io_array[alive_cmds++];
-            iov.iov_base = (void*)buf_cursor;
-            iov.iov_len = sz;
-        }
-        if (1 == alive_cmds) [[likely]]
-            io_addr = logical_off;
+        auto& acc = sub_cmds[stripe_off];
+        if (!acc.alive_cmds) acc.io_addr = logical_off;
+        acc.io_array[acc.alive_cmds++] = {buf_cursor, sz};
 
-        // Last sub_cmd for this device, issue now
+        // Dispatch once the remaining bytes fit within a single (N-1)-stripe remainder,
+        // guaranteeing this stripe cannot accumulate more iovecs in the same call.
         if ((_stride_width - _stripe_size) >= (len - off)) {
             sub_cmd_t const new_sub_cmd = sub_cmd + (!retry ? (uint16_t)stripe_off : 0);
-            DEBUG_ASSERT_LE(alive_cmds, UINT32_MAX) // LCOV_EXCL_LINE
-            auto res = func(stripe_off, new_sub_cmd, io_array.data(), alive_cmds, io_addr);
+            auto res = func(stripe_off, new_sub_cmd, acc.io_array.data(), acc.alive_cmds, acc.io_addr);
             if (!res) return res;
             cnt += res.value();
         }
--- a/src/raid/raid0/raid0.cpp
+++ b/src/raid/raid0/raid0.cpp
@@ -175,9 +175,24 @@ io_result Raid0Disk::handle_discard(ublksrv_queue const* q, ublk_io_data const* 
 //  stripe boundaries and even wrap around several strides. This routine handles this calculation and calls
 //  the given routine `func` for each stripe that it has collected scatter (struct iovec) operations for.
 io_result Raid0Disk::__distribute(iovec* iovecs, uint64_t addr, auto&& func, bool retry, sub_cmd_t sub_cmd) const {
-    // Each thread has a 2-dimensional block of iovecs that we can split into
+    // Each thread has a 2-dimensional block of iovecs that we can split into.
     thread_local auto sub_cmds =
         std::array< std::tuple< uint64_t, uint32_t, std::array< iovec, 16 > >, _max_stripe_cnt >();
+    // Reset only the stripes we touch (cheaper than zeroing all _max_stripe_cnt on every call).
+    // A previous call that returned early on error leaves non-zero alive_cmds behind; stale iov_base
+    // pointers in those entries would corrupt the next I/O on this thread if not cleared.
+    static_assert(_max_stripe_cnt <= 64, "dirty_mask requires _max_stripe_cnt <= 64");
+    uint64_t dirty_mask{0};
+    struct DirtyGuard {
+        decltype(sub_cmds)& cmds;
+        uint64_t& mask;
+        ~DirtyGuard() noexcept {
+            while (mask) {
+                std::get< 1 >(cmds[std::countr_zero(mask)]) = 0;
+                mask &= mask - 1;
+            }
+        }
+    } _guard{sub_cmds, dirty_mask};
 
     // Special case for single device
     if (1 == _stripe_array.size()) return func(0, sub_cmd, iovecs, 1, addr);
@@ -205,6 +220,7 @@ io_result Raid0Disk::__distribute(iovec* iovecs, uint64_t addr, auto&& func, boo
             if (stripe_off != ((sub_cmd >> device->route_size()) & route_mask)) continue;
         }
 
+        dirty_mask |= 1ULL << stripe_off;
         auto& [io_addr, alive_cmds, io_array] = sub_cmds[stripe_off];
         { // Fillout iovec
             auto& iov = io_array[alive_cmds++];
--- a/src/raid/raid0/raid0.cpp
+++ b/src/raid/raid0/raid0.cpp
@@ -235,8 +235,6 @@ io_result Raid0Disk::__distribute(iovec* iovecs, uint64_t addr, auto&& func, boo
             sub_cmd_t const new_sub_cmd = sub_cmd + (!retry ? (uint16_t)stripe_off : 0);
             DEBUG_ASSERT_LE(alive_cmds, UINT32_MAX) // LCOV_EXCL_LINE
             auto res = func(stripe_off, new_sub_cmd, io_array.data(), alive_cmds, io_addr);
-            // Set this back to zero so the next command can reuse
-            alive_cmds = 0;
             if (!res) return res;
             cnt += res.value();
         }
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -277,6 +277,12 @@ Raid1DiskImpl::~Raid1DiskImpl() {
 }
 
 std::list< int > Raid1DiskImpl::open_for_uring(int const iouring_device_start) {
+    // Called once per queue thread before I/O begins; count queues for multi-queue idle tracking.
+    ++_nr_hw_queues;
+
+    // Only initialize on the first call (first queue thread)
+    if (_nr_hw_queues > 1) return {};
+
     // Now that we're up and the target wants to begin I/O let's unpause our resync task, there are no i/o threads
     // yet so we can continue to access _device_a and b directly
     auto fds = (_device_a->disk->open_for_uring(iouring_device_start));
@@ -970,10 +976,18 @@ void Raid1DiskImpl::on_io_complete(ublk_io_data const* data, sub_cmd_t sub_cmd, 
 
 void Raid1DiskImpl::idle_transition(ublksrv_queue const*, bool enter) noexcept {
     if (!enter) {
+        _idle_queue_count.fetch_sub(1, std::memory_order_acq_rel);
+        auto lk = std::unique_lock{_idle_probe_lock};
         _idle_probe_a.stop();
         _idle_probe_b.stop();
         return;
     }
+
+    // Start probes only when all queue threads are idle (_nr_hw_queues == 0 treated as 1 for
+    // backward compat with tests that skip open_for_uring).
+    auto const nr = _nr_hw_queues ? _nr_hw_queues : uint16_t{1};
+    auto const prev = _idle_queue_count.fetch_add(1, std::memory_order_acq_rel);
+    if (prev + 1 < nr) return;
 
     auto const state = __capture_route_state();
     if (state.is_degraded) return; // Resync task handles avail probing in degraded mode
@@ -989,7 +1003,8 @@ void Raid1DiskImpl::idle_transition(ublksrv_queue const*, bool enter) noexcept {
     immediate_probe(state.active_dev);
     immediate_probe(state.backup_dev);
 
-    // Periodic probe: both directions (recovery + new failures), stopped when idle exits
+    // Periodic probe: both directions (recovery + new failures), stopped when any queue exits idle
+    auto lk = std::unique_lock{_idle_probe_lock};
     _idle_probe_a.launch(state.active_dev, reserved_size);
     _idle_probe_b.launch(state.backup_dev, reserved_size);
 }
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -283,7 +283,7 @@ std::list< int > Raid1DiskImpl::open_for_uring(int const iouring_device_start) {
     fds.splice(fds.end(), _device_b->disk->open_for_uring(iouring_device_start + fds.size()));
 
     // Enable resync only on the first call (first queue thread).
-    if (++_nr_hw_queues == 1) toggle_resync(true);
+    if (_nr_hw_queues.fetch_add(1, std::memory_order_acq_rel) == 0) toggle_resync(true);
 
     return fds;
 }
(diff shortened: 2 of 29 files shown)


src/raid/raid1/raid1.cpp
```diff
@@ -970,11 +997,17 @@ void Raid1DiskImpl::on_io_complete(ublk_io_data const* data, sub_cmd_t sub_cmd,
 
 void Raid1DiskImpl::idle_transition(ublksrv_queue const*, bool enter) noexcept {
     if (!enter) {
+        _idle_queue_count.fetch_sub(1, std::memory_order_acq_rel);
+        auto lk = std::unique_lock{_idle_probe_lock};
```
Ilyin-Ilya, line 1005: **Bug: `_idle_queue_count` can underflow (uint16_t wrap → 65535)**

The exit path decrements unconditionally:
```cpp
_idle_queue_count.fetch_sub(1, std::memory_order_acq_rel);
```
If `idle_transition(q, false)` fires without a preceding `idle_transition(q, true)` — e.g. during driver teardown, queue reset, or a spurious exit event — the `uint16_t` wraps to 65535. Subsequent enter calls will always find `prev + 1 < 65535`, so the probe **never launches again** for the lifetime of the array, permanently suppressing availability recovery.

Suggested fix: guard the decrement, or use a saturating subtract:
```cpp
// Only decrement if count > 0
uint16_t prev = _idle_queue_count.load(std::memory_order_relaxed);
while (prev > 0) {
    if (_idle_queue_count.compare_exchange_weak(prev, prev - 1,
            std::memory_order_acq_rel, std::memory_order_relaxed))
        break;
}
```
Or at minimum add a debug assertion: `DASSERT(_idle_queue_count.load() > 0, "exit without enter");`

src/raid/raid1/raid1.cpp
```diff
@@ -970,11 +997,17 @@ void Raid1DiskImpl::on_io_complete(ublk_io_data const* data, sub_cmd_t sub_cmd,
...
         _idle_probe_b.stop();
         return;
     }
 
+    // Start probes only when all queue threads are idle.
+    auto const prev = _idle_queue_count.fetch_add(1, std::memory_order_acq_rel);
+    if (prev + 1 < _nr_hw_queues.load(std::memory_order_acquire)) return;
+
     auto const state = __capture_route_state();
     if (state.is_degraded) return; // Resync task handles avail probing in degraded mode
```
Ilyin-Ilya, line 1019: **Race: route state captured before `_idle_probe_lock` is acquired**

The enter path captures state outside the lock:
```cpp
auto const prev = _idle_queue_count.fetch_add(1, std::memory_order_acq_rel);
if (prev + 1 < _nr_hw_queues.load(std::memory_order_acquire)) return;

auto const state = __capture_route_state();  // ← outside the lock
if (state.is_degraded) return;
// ...
auto lk = std::unique_lock{_idle_probe_lock};  // ← lock acquired here
_idle_probe_a.launch(state.active_dev, reserved_size);  // ← stale state used
```

Between `__capture_route_state()` and acquiring `_idle_probe_lock`, `swap_device()` can run: it acquires `_idle_probe_lock`, calls `stop()` on both probes (with the outgoing device), then releases the lock. `idle_transition` then acquires the lock and calls `launch()` with `state.active_dev` pointing to the **outgoing device that was just swapped out**. The probe then holds a `shared_ptr` to the dead device and probes it indefinitely.

Suggested fix: re-capture state inside the locked section, or re-verify state has not changed after acquiring the lock:
```cpp
auto lk = std::unique_lock{_idle_probe_lock};
auto const state = __capture_route_state();  // re-capture under lock
if (state.is_degraded) return;
// ... compute reserved_size ...
_idle_probe_a.launch(state.active_dev, reserved_size);
_idle_probe_b.launch(state.backup_dev, reserved_size);
```
The `immediate_probe` calls before the lock could stay outside (they are fire-and-forget), but the `launch()` calls must use state captured under the lock.

src/raid/raid1/raid1.cpp
```diff
@@ -661,7 +687,7 @@ io_result Raid1DiskImpl::__handle_async_retry(sub_cmd_t sub_cmd, uint64_t addr,
     // Instead, inject a synthetic async_result carrying `len` into _pending_results so it is accumulated
     // into ret_val via the normal process_result path on the next collect_async cycle. Return +1 to
     // signal exactly one more sub_cmd pending.
-    _pending_results[q].emplace_back(async_result{async_data, sub_cmd, static_cast< int >(len)});
+    _pending_results.at(q).emplace_back(async_result{async_data, sub_cmd, static_cast< int >(len)});
     if (q) {
```
Ilyin-Ilya, line 693: **Latent crash: `_pending_results.at(q)` throws when `q == nullptr`**

`open_for_uring` guards insertion with `if (q)`, so `nullptr` is never inserted into `_pending_results`. However, `__handle_async_retry` now uses `.at(q)` unconditionally:
```cpp
_pending_results.at(q).emplace_back(...);  // throws std::out_of_range when q == nullptr
```

Meanwhile, `collect_async` uses the safe `.find(q)` path. The two call sites have inconsistent assumptions about map membership.

In production, `q` should always be non-null on the I/O path, so this is latent. But if any test or future code path reaches `__handle_async_retry` with a null queue (the mock infrastructure supports `nullptr` queues via the `if (q)` guard in `open_for_uring`), it throws and kills the process.

Suggested fix: use the same `.find()` pattern as `collect_async`, or add an assertion:
```cpp
auto it = _pending_results.find(q);
DASSERT(it != _pending_results.end(), "queue not registered");
it->second.emplace_back(...);
```

src/target/ublkpp_tgt.cpp
```diff
@@ -442,8 +441,10 @@ static void idle_transition(ublksrv_queue const* q, bool enter) {
     auto device = reinterpret_cast< UblkDisk* >(q->dev->tgt.tgt_data);
     device->idle_transition(q, enter);
 }
-static int init_queue(const struct ublksrv_queue*, void**) {
+static int init_queue(const struct ublksrv_queue* q, void**) {
     TLOGD("Init Queue")
+    auto device = reinterpret_cast< UblkDisk* >(q->dev->tgt.tgt_data);
+    device->open_for_uring(q, 0);
     return 0;
```
Ilyin-Ilya, line 450: **Silent regression: FD list from `open_for_uring` is discarded in `init_queue`**

```cpp
device->open_for_uring(q, 0);  // return value discarded
```

Any disk that returns non-empty FDs from `open_for_uring` will silently fail to register them with `io_uring`, and subsequent fixed-file I/O for those FDs will crash or produce errors. The comment above accurately describes the constraint but provides no enforcement:

> "if future disks export non-empty FDs they must be registered here"

The return value should be captured and either registered or asserted-empty:
```cpp
auto fds = device->open_for_uring(q, 0);
// FDs returned here must be registered via io_uring_register_files before I/O begins.
// Currently all disk types return empty; assert so we catch regressions immediately.
RELEASE_ASSERT(fds.empty(), "open_for_uring returned FDs that are not being registered");
```

This is particularly important because the original `init_tgt` path that did register FDs was removed by this PR, so any future disk returning FDs will have no registration path at all.

src/raid/raid1/raid1_impl.hpp
```diff
@@ -49,13 +49,18 @@ class Raid1DiskImpl : public UblkDisk {
     std::map< ublksrv_queue const*, std::list< async_result > > _pending_results;
 
     // Active Re-Sync Task
-    bool _resync_enabled{true};
+    std::atomic< bool > _resync_enabled{true};
     std::shared_ptr< Raid1ResyncTask > _resync_task;
 
     // Ensure exclusivity in __swap_device
     std::mutex _swap_lock;
```
Ilyin-Ilya: **Stale comment: `_swap_lock` now guards two unrelated concerns**

The comment still says "Ensure exclusivity in `__swap_device`":
```cpp
// Ensure exclusivity in __swap_device
std::mutex _swap_lock;
```

But this PR adds a second use: `open_for_uring` acquires `_swap_lock` to protect `_pending_results.emplace()`. These are semantically unrelated concerns — device-swap exclusivity vs. per-queue map initialization — sharing one lock makes the invariants harder to reason about and the comment actively misleads.

Two options:
1. Update the comment to enumerate all protected resources:
```cpp
// Guards: (1) __swap_device — serializes concurrent swap_device() callers on _device_a/_device_b mutations.
//         (2) _pending_results — serializes open_for_uring() insertions across concurrent queue threads.
std::mutex _swap_lock;
```
2. Introduce a dedicated `_pending_results_lock` (simpler invariants, avoids holding swap-scoped lock during init).

src/raid/raid0/raid0.cpp
```diff
@@ -175,9 +175,24 @@ io_result Raid0Disk::handle_discard(ublksrv_queue const* q, ublk_io_data const*
...
+    struct DirtyGuard {
+        decltype(sub_cmds)& cmds;
+        uint64_t& mask;
+        ~DirtyGuard() noexcept {
+            while (mask) {
+                std::get< 1 >(cmds[std::countr_zero(mask)]) = 0;
+                mask &= mask - 1;
+            }
+        }
+    } _guard{sub_cmds, dirty_mask};
```
Ilyin-Ilya, line 201: **Nit: local variable `_guard` uses member-style `_` prefix**

```cpp
} _guard{sub_cmds, dirty_mask};
```

`_guard` is a local variable but follows the `_snake_case` naming convention reserved for class members (per `CLAUDE.md`). Using the wrong prefix is a readability signal that this is a member field — a maintainer scanning the function will misread its lifetime and scope.

Rename to `guard` or `dirty_guard`:
```cpp
} guard{sub_cmds, dirty_mask};
```

Ilyin-Ilya (commented):
## Review Summary

This PR tackles a real and important set of bugs — multi-queue safety for the idle probe, the `operator[]`-on-concurrent-access bug in `_pending_results`, the `memset`/atomic UB in `SuperBitmap::clear_all()`, and the RAID-0 `alive_cmds` stale-state corruption. The overall approach is sound and the new concurrency tests (`MultiQueueIdleConcurrent` with `std::barrier`) are a genuine improvement.

That said, the new idle-probe path introduces two new races of its own, and several secondary issues are worth addressing before merge.

---

### Critical — likely to produce incorrect behavior in production

**1. `_idle_queue_count` underflow on spurious exit (`src/raid/raid1/raid1.cpp`, `idle_transition` exit path)**
The decrement is unconditional. If `idle_transition(q, false)` fires without a matching enter — driver teardown, reset, or a spurious event — the `uint16_t` wraps to 65535. Every subsequent enter finds `prev + 1 < 65535` and returns early; probes never launch again. See inline comment for a saturating-CAS fix.

**2. Stale route state used after lock acquisition in `idle_transition` enter path (`src/raid/raid1/raid1.cpp`)**
`__capture_route_state()` is called before acquiring `_idle_probe_lock`. Between that call and the lock, `swap_device()` can win the lock, stop the probes, and release the lock. `idle_transition` then acquires the lock and calls `launch()` with `state.active_dev` pointing to the outgoing device. The fix is to re-capture state inside the locked section before calling `launch()`. See inline comment.

---

### Major — correctness risk or silent regression

**3. `_pending_results.at(q)` throws on `q == nullptr` (`src/raid/raid1/raid1.cpp`, `__handle_async_retry`)**
`open_for_uring` guards insertion with `if (q)`, so `nullptr` is never in the map. `__handle_async_retry` uses `.at(q)` unconditionally and throws on a miss; `collect_async` uses the safe `.find()` path. These two call sites are inconsistent. In production `q` should be non-null on the I/O path, but the inconsistency is fragile. Suggest aligning both on `.find()` + assertion. See inline comment.

**4. FD return value silently discarded in `init_queue` (`src/target/ublkpp_tgt.cpp`)**
`device->open_for_uring(q, 0)` is called with its result dropped. The original `init_tgt` registration loop has been removed entirely. Any disk that returns non-empty FDs will silently fail to register them, causing crashes on fixed-file I/O with no diagnostic. The comment is accurate but unenforced — at minimum, assert that the returned list is empty. See inline comment.

---

### Minor — code quality / maintainability

**5. `_swap_lock` comment is stale (`src/raid/raid1/raid1_impl.hpp`)**
Comment says "Ensure exclusivity in `__swap_device`" but the lock now also serializes `_pending_results.emplace()` insertions in `open_for_uring`. Two options: update the comment to enumerate both guarded resources, or introduce a dedicated `_pending_results_lock`. See inline comment.

**6. Local variable `_guard` uses member-style `_` prefix (`src/raid/raid0/raid0.cpp`)**
`DirtyGuard _guard{sub_cmds, dirty_mask}` is a local variable using the member-field naming convention. Should be `guard` or `dirty_guard`. See inline comment.

**7. Missing tests for the two new races (items 1 and 2 above)**
`MultiQueueIdleSequential` exercises the balanced enter/exit path but not the exit-without-enter case. A `SwapDeviceWhileIdleTransitioning` test that races `swap_device()` against `idle_transition()` concurrently would be the authoritative TSAN guard for the stale-state race (item 2).

**8. `open_for_uring` signature change is a breaking virtual API change not noted in CHANGELOG**
The CHANGELOG entry covers the behavioral fix but does not mention the signature change. Any out-of-tree subclass of `UblkDisk` that overrides `open_for_uring` will silently fall back to the no-op base implementation with no compile error (it no longer matches the virtual).

---

### Positive aspects

- The `_idle_probe_lock` around `swap_device`'s `stop()` calls correctly closes the original race from #193.
- `SuperBitmap::clear_all()` atomic byte-store fix is correct and well-commented.
- The `DirtyGuard` RAII approach for RAID-0 `alive_cmds` cleanup is clean and cheaper than zeroing all 64 entries on every call.
- `_resync_enabled` upgrade to `std::atomic<bool>` and the `_resync_enabled.load()` call at the use sites is a good hygiene fix.
- `MockUblksrv` multi-queue support is a meaningful test infrastructure improvement.
- `std::barrier` synchronization in `ConcurrentReadWhileWrite` and `MultiQueueIdleConcurrent` correctly eliminates the scheduling-dependent false-negative that was causing CI flakes.

---

*Review performed by Claude Code (claude-sonnet-4-6)*

src/raid/raid1/raid1.cpp
szmyd: Good catch on the analysis, but the scenario can't happen in practice due to ublksrv's idle-transition contract.

`idle_transition()` is called by ublksrv's queue-thread event loop, which maintains a strict per-queue boolean: it calls `idle_transition(q, true)` exactly once when the queue first drains, and `idle_transition(q, false)` exactly once when the next I/O arrives. There is no code path in ublksrv that fires a false exit without a matching enter — it's not a poll-based API where spurious wakeups occur.

During teardown, ublksrv stops I/O dispatch before calling any cleanup hooks, so `idle_transition` is not called at all on the shutdown path.

The consequence of underflow, *if* it did happen, would be benign in any case: the probe simply wouldn't launch until the counter recovered (next full enter cycle). It wouldn't cause data corruption or a stuck array — the probe is an optimization that recovers routing to a previously-unavailable device; the array is still fully operational without it. A CAS loop in a `noexcept` hot path for a scenario the driver contract excludes is more complexity than the risk warrants.

Agreeing with the `DASSERT` suggestion though — that's cheap and would catch any future ublksrv behaviour change that violated the assumption during testing.

src/raid/raid1/tests/concurrency/multi_queue_idle.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    // Simulate k_queues queue threads (sets _nr_hw_queues = k_queues)
+    for (int i = 0; i < k_queues; ++i)
+        raid_device.open_for_uring(nullptr, 0);
+
+    // Allow any number of sync_iov calls in case the periodic probe fires unexpectedly
+    EXPECT_CALL(*device_a, sync_iov(::testing::_, ::testing::_, ::testing::_, ::testing::_))
+        .Times(::testing::AnyNumber())
+        .WillRepeatedly([](uint8_t, iovec* iovecs, uint32_t, off_t) -> ublkpp::io_result {
+            return static_cast< int >(iovecs->iov_len);
```
Ilyin-Ilya: The `sync_iov` mocks in `MultiQueueIdleConcurrent` and `MultiQueueIdleRapidToggle` return `iovecs->iov_len` without zeroing the buffer — the same pattern the PR fixes in `ResyncRelaunchAfterComplete`, `ConcurrentLaunch`, and `StopDuringUnavailWait` via `sync_iov_zero_on_read()`. Uninitialized `posix_memalign` memory can look like a valid superblock with a mismatched UUID, causing spurious load failures on sanitize builds.

Suggested fix:
```cpp
.WillRepeatedly(sync_iov_zero_on_read())
```

src/tests/mock_ublksrv/mock_ublksrv.cpp
```diff
@@ -28,19 +28,23 @@ MockUblksrv::MockUblksrv(std::shared_ptr< UblkDisk > disk, int q_depth) :
...
 
-    // Register any FDs the disk exports (Raid0/Raid1 forward from FsDisks;
-    // FSDisk currently returns empty since fixed-file support is disabled)
-    for (auto const fd : _disk->open_for_uring(_dev.tgt.nr_fds)) {
-        if (_dev.tgt.nr_fds < UBLKSRV_TGT_MAX_FDS) _dev.tgt.fds[_dev.tgt.nr_fds++] = fd;
+    // Simulate init_queue: call open_for_uring once per queue thread so the disk can count queues
+    // and perform per-queue initialization (e.g. Raid1DiskImpl sets _nr_hw_queues and enables resync).
+    for (int qi = 0; qi < nr_queues; ++qi) {
+        for (auto const fd : _disk->open_for_uring(&_queues[qi], _dev.tgt.nr_fds)) {
+            if (_dev.tgt.nr_fds < UBLKSRV_TGT_MAX_FDS) _dev.tgt.fds[_dev.tgt.nr_fds++] = fd;
```
Ilyin-Ilya: `_queues[qi]` fields (`q_id`, `q_depth`, `ring_ptr`, `dev`) are populated in the second loop, but `open_for_uring(&_queues[qi], ...)` is called here in the first loop — so the pointer passed to `open_for_uring` points to a zero-initialized struct. Any disk implementation that dereferences queue fields inside `open_for_uring` (e.g., to call `ublksrv_queue_send_event`) will read zeroes instead of the real values.

Suggested fix: swap the loop order — populate queue struct fields first, then call `open_for_uring`.

src/raid/raid1/tests/concurrency/multi_queue_idle.cpp
```diff
@@ -0,0 +1,132 @@
...
+    auto device_b = CREATE_DISK_B(TestParams{.capacity = Gi});
+    auto raid_device = ublkpp::Raid1Disk(boost::uuids::string_generator()(test_uuid), device_a, device_b);
+
+    constexpr int k_queues = 4;
+    constexpr int k_iters = 50;
+
+    // Simulate k_queues queue threads (sets _nr_hw_queues = k_queues)
+    for (int i = 0; i < k_queues; ++i)
+        raid_device.open_for_uring(nullptr, 0);
+
```
Ilyin-Ilya, line 68: `open_for_uring` is called with `nullptr` here, so the `if (q)` guard in `Raid1DiskImpl::open_for_uring` skips the `_pending_results.emplace` call. The `_pending_results` pre-population path — one of the three bugs fixed by this PR — is never exercised by any of the new `MultiQueue*` tests.

Consider passing a real queue pointer (e.g., from a `MockUblksrv` with `nr_queues=2`) so the new lock and emplace path is covered.

src/raid/raid1/raid1_impl.hpp
```diff
@@ -49,13 +49,19 @@ class Raid1DiskImpl : public UblkDisk {
...
 
-    // Ensure exclusivity in __swap_device
-    std::mutex _swap_lock;
+    // Guards: (1) swap_device() — serializes concurrent callers on _device_a/_device_b mutations.
+    //         (2) _pending_results — serializes open_for_uring() insertions across queue threads.
+    std::mutex _ctrl_lock;
+
+    // Multi-queue idle tracking: probe starts when all queues are idle, stops on any active transition
+    std::atomic_uint16_t _nr_hw_queues{0};
+    std::atomic_uint16_t _idle_queue_count{0};
```
Ilyin-Ilya, line 62: `_idle_queue_count` is `uint16_t`. In release builds `DEBUG_ASSERT` is stripped, so a spurious `idle_transition(false)` call (ublksrv contract violation) wraps the counter from 0 to 65535 — permanently suppressing all future probes with no log message and no way to recover short of restarting.

Consider a release-mode guard:
```cpp
if (_idle_queue_count.load(std::memory_order_relaxed) == 0) {
    RLOGE("idle_transition exit without matching enter");
    return;
}
_idle_queue_count.fetch_sub(1, std::memory_order_acq_rel);
```

src/raid/raid1/raid1.cpp
```diff
@@ -276,27 +276,46 @@ Raid1DiskImpl::~Raid1DiskImpl() {
...
-    auto fds = (_device_a->disk->open_for_uring(iouring_device_start));
-    fds.splice(fds.end(), _device_b->disk->open_for_uring(iouring_device_start + fds.size()));
-
-    // I/O will start comming in now, enable resync
-    toggle_resync(true);
+std::list< int > Raid1DiskImpl::open_for_uring(ublksrv_queue const* q, int const iouring_device_start) {
+    // Called once per queue thread before I/O begins; count queues for multi-queue idle tracking.
+    // Always collect FDs from child disks — each queue thread may need its own set.
+    auto fds = _device_a->disk->open_for_uring(q, iouring_device_start);
+    fds.splice(fds.end(), _device_b->disk->open_for_uring(q, iouring_device_start + fds.size()));
```
Ilyin-Ilya, line 283: When `_nr_hw_queues == 0` (its initial value, before any `open_for_uring` call), this gate evaluates as `prev + 1 < 0` — which for `uint16_t` is `1 < 0` = false — so the probe fires unconditionally. Existing tests that call `idle_transition(true)` without first calling `open_for_uring` (e.g., `IdleTransitionEnter`) rely on this unsigned underflow silently.

This works, but is fragile: if `_nr_hw_queues` were ever initialized to 1 (e.g., to represent a default single-queue assumption), those tests would break without any obvious connection to this line. Consider adding a comment documenting the zero-queue contract:
```cpp
// When _nr_hw_queues == 0 (no open_for_uring call yet), prev+1 < 0 is false for uint16_t,
// so the probe fires unconditionally — preserving backward compatibility with single-queue callers
// that do not call open_for_uring.
if (prev + 1 < _nr_hw_queues.load(std::memory_order_acquire)) return;
```

src/raid/raid1/tests/simple/idle_and_collect.cpp
```diff
@@ -23,6 +23,9 @@ TEST(Raid1, IdleTransitionExit) {
     auto raid_device = ublkpp::Raid1Disk(boost::uuids::string_generator()(test_uuid), device_a, device_b);
 
     // Should not crash - manages internal resync state
+    // enter must precede exit to satisfy the ublksrv contract (both devices are available
+    // so the immediate probe inside idle_transition(true) is a no-op)
+    raid_device.idle_transition(nullptr, true);
     raid_device.idle_transition(nullptr, false);
```
Ilyin-Ilya, line 29: The test was updated to call `idle_transition(true)` before `idle_transition(false)` to satisfy the new `DEBUG_ASSERT`, but it is still named `IdleTransitionExit` and the outer comment still reads "Should not crash — manages internal resync state". The name and comment now misrepresent the test's scope.

Suggested rename: `IdleTransitionRoundTrip` or `IdleTransitionEnterAndExit`.

Ilyin-Ilya (approved):

PR #206, opened 2026-04-24
fix(raid1): eliminate enqueue/dequeue pause race via atomic_status_counter

szmyd:
## Summary

Fixes two independent logical races in the RAID1 resync pause/resume protocol.

### Race 1: dequeue/resume race (primary fix)

- **Root cause**: `_resync_state` and `_outstanding_writes` were separate atomics, creating a TOCTOU window where `dequeue_write()` decremented the counter to zero and then called `__resume()` (CAS PAUSE→ACTIVE) as two distinct steps. A concurrent `enqueue_write()` in that window saw `count==0`, called `__pause()`, found state still PAUSE, and early-exited — trusting an existing pause that was about to be cleared. `__resume()` then fired, state became ACTIVE, and resync ran concurrently with the in-flight write.
- **Fix**: Replace the two fields with `sisl::atomic_status_counter<resync_state>`, which packs both into a single 64-bit lock-free CAS word. `dequeue_write()` now calls `dec_xchng_status_ifz(PAUSE, ACTIVE)`, making the decrement and the conditional state transition one indivisible operation. The `(count==0, status==PAUSE)` state is never observable — when count hits zero, status atomically becomes ACTIVE, so `__pause()` can never early-exit on a stale pause.

### Race 2: enqueue/pause race (pre-existing, surfaced by the new test)

- **Root cause**: `enqueue_write()` only called `__pause()` when the counter was zero (`was_zero==true`). If Thread A incremented the counter 0→1 but had not yet completed `__pause()`, Thread B would see `was_zero=false`, skip `__pause()`, and proceed to submit its write while resync was still ACTIVE — allowing resync to overwrite the write with stale data from the clean mirror.
- **Fix**: Always call `__pause()` after incrementing the counter. When state is already PAUSE the call exits in O(1) (one failing CAS + early-exit); the spin cost is only paid by threads that race Thread A's initial pause establishment.

### Other changes

- **Documentation**: Comments at the `_state_and_writes` declaration and across `enqueue_write()`, `dequeue_write()`, and `stop()` document the ordering contract, the lock-free guarantee, and the source-state invariant relied on by `__yield`'s `DEBUG_ASSERT_NE`.
- **Cleanup**: `__resume()` removed (no longer needed); `__cas_state_weak` + `__cas_state_strong` consolidated into a single `__cas_state` (were identical after the refactor); unreachable STOPPING check in `__yield` Phase 1 replaced with `DEBUG_ASSERT` + `LCOV_EXCL_LINE`; `stop()` return type changed from `uint32_t` to `void` (the outstanding-write count at join time is not a meaningful signal — callers must drain I/O before calling `stop()`).
- **C++23 `<bit>` standardisation** (`bitmap.cpp`, `super_bitmap.cpp`): replaced five GCC extension calls with their standard equivalents — `__builtin_popcountll` → `std::popcount` (dirty chunk accounting in `clean_region` and `dirty_region`), `__builtin_clzl` → `std::countl_zero` (first-dirty-bit scan in `next_dirty`), `__builtin_ctz` × 2 → `std::countr_zero` (first-set-byte scan in `SuperBitmap::next_set_bit`). Also fixed a latent portability hazard: `auto word = 0UL` (32-bit `unsigned long` on LLP64) changed to `uint64_t word = 0` so `std::countl_zero` always operates on 64 bits.
- **Lock-free atomic**: `resync_state` underlying type changed from `uint8_t` to `uint32_t` so `_status_counter` is 8 bytes (int32_t counter + uint32_t status), natively atomic via `CMPXCHG8B` on x86-64 with no libatomic calls. A `static_assert` enforces this.
- **Build**: `libatomic` declared as a Conan `system_libs` entry on Linux and added explicitly to the fio engine `.so` link as a safety dependency (static libs don't propagate system libs to shared library links); the `_state_and_writes` field itself no longer requires it.
- **Refactor**: `_read_route_cache` changed from `std::atomic_uint8_t` to `std::atomic<raid1::read_route>` — eliminates `static_cast<uint8_t>` noise at every CAS/load/store site and removes the `uint8_t`-overload helpers that existed solely to avoid those casts. Memory ordering at every site is unchanged.

## Test plan

- [x] Regression test added: `Raid1Concurrency.EnqueueDequeueRace` — 4 threads × 200 enqueue/dequeue pairs against a slow-copy resync with an atomic sentinel ensuring resync has begun before I/O threads start; asserts no resync write overlaps an in-flight I/O write (exposes both Race 1 and Race 2)
- [x] All existing concurrency tests pass
- [x] TSAN build passes (`-o sanitize=thread`)
- [x] ASan build passes (`-o sanitize=address`)
- [x] Debug build passes
- [x] fio functional tests pass

> **Note on Race 1 detectability**: The window is ~2–5 CPU instructions between `fetch_sub` and the former `__resume()` call — too narrow for probabilistic test detection, and a logical race on atomics (not a C++ data race), so TSAN cannot flag it. Correctness is an analytical argument. Race 2 is reliably detected by the regression test under ASan/TSan (instrumentation widens the spin window); a plain Debug run is not a regression guard for either race.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## 0.22.1
-- raid1: Fix enqueue_write/dequeue_write pause race — `_resync_state` and `_outstanding_writes` are now packed into a single `sisl::atomic_status_counter` so the decrement-and-resume is one indivisible CAS, closing the window where resync could run concurrently with an in-flight write (SDSTOR-21927)
+- raid1: Fix dequeue/resume race — `_resync_state` and `_outstanding_writes` are now packed into a single `sisl::atomic_status_counter` so the counter decrement and PAUSE→ACTIVE transition are one indivisible CAS; `__resume()` is removed (SDSTOR-21927)
+- raid1: Fix enqueue/pause race — `enqueue_write()` now always calls `__pause()` on every enqueue, not only the first; previously a concurrent second enqueuer could skip `__pause()` while the first was still establishing it, allowing resync to overwrite an in-flight write with stale data
 - raid1: Replace GCC `__builtin_popcount`/`__builtin_clz`/`__builtin_ctz` with C++23 `std::popcount`/`std::countl_zero`/`std::countr_zero`
 - build: `libatomic` is now declared as a Conan system lib on Linux — propagated automatically to consumers, no downstream changes required
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.22.1
+- raid1: Fix enqueue_write/dequeue_write pause race — `_resync_state` and `_outstanding_writes` are now packed into a single `sisl::atomic_status_counter` so the decrement-and-resume is one indivisible CAS, closing the window where resync could run concurrently with an in-flight write (SDSTOR-21927)
+- raid1: Replace GCC `__builtin_popcount`/`__builtin_clz`/`__builtin_ctz` with C++23 `std::popcount`/`std::countl_zero`/`std::countr_zero`
+- build: `libatomic` is now declared as a Conan system lib on Linux — propagated automatically to consumers, no downstream changes required
+
 ## 0.22.0
 - raid1: Fix multi-queue idle probe race conditions — probes now start only when all queues are idle, mutex serializes concurrent launch/stop calls, `open_for_uring` counts queue threads for accurate `nr_hw_queues`
 - **Breaking**: `UblkDisk::open_for_uring` signature changed from `(int)` to `(ublksrv_queue const*, int)` — out-of-tree subclasses must update their override
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=2.0"
 
 class UBlkPPConan(ConanFile):
     name = "ublkpp"
-    version = "0.22.0"
+    version = "0.22.1"
 
     homepage = "https://github.com/szmyd/ublkpp"
     description = "A UBlk library for CPP application"
@@ -146,6 +146,8 @@ class UBlkPPConan(ConanFile):
 
     def package_info(self):
         self.cpp_info.requires = ["sisl::cache", "isa-l::isa-l", "ublksrv::ublksrv"]
+        if self.settings.os == "Linux":
+            self.cpp_info.system_libs = ["atomic"]
         if (self.options.get_safe("iscsi")):
             self.cpp_info.requires.extend(["libiscsi::libiscsi"])
         if (self.options.get_safe("homeblocks")):
--- a/conanfile.py
+++ b/conanfile.py
@@ -47,7 +47,7 @@ class UBlkPPConan(ConanFile):
                         )
 
     def _min_cppstd(self):
-        return 20
+        return 23
 
     def validate(self):
         if self.settings.compiler.get_safe("cppstd"):
--- a/src/raid/raid1/bitmap.cpp
+++ b/src/raid/raid1/bitmap.cpp
@@ -348,7 +348,7 @@ std::pair< uint64_t, uint32_t > Bitmap::next_dirty() noexcept {
         logical_off = static_cast< uint64_t >(_page_width) * pg_off;
 
         // Find the first dirty word
-        auto word = 0UL;
+        uint64_t word = 0;
         for (auto word_off = 0U; (k_page_size / sizeof(word_t)) > word_off; ++word_off) {
             word = be64toh((page + word_off)->load(std::memory_order_relaxed));
             if (0 == word) continue;
--- a/src/raid/raid1/bitmap.cpp
+++ b/src/raid/raid1/bitmap.cpp
@@ -1,5 +1,6 @@
 #include "bitmap.hpp"
 
+#include <bit>
 #include <isa-l/mem_routines.h>
 #include <ublk_cmd.h>
 
@@ -305,9 +306,10 @@ std::tuple< Bitmap::word_t*, uint32_t, uint32_t > Bitmap::clean_region(uint64_t 
                                                              : (((uint64_t)0b1 << bits_to_write) - 1)
                                                  << (shift_offset - (bits_to_write - 1)));
         auto old_word = cur_word->fetch_and(clear_mask, std::memory_order_relaxed);
-        _dirty_chunks_est.fetch_sub(std::min(_dirty_chunks_est.load(std::memory_order_relaxed),
-                                             (uint64_t)__builtin_popcountll(old_word xor (old_word & clear_mask))),
-                                    std::memory_order_relaxed);
+        _dirty_chunks_est.fetch_sub(
+            std::min(_dirty_chunks_est.load(std::memory_order_relaxed),
+                     static_cast< uint64_t >(std::popcount(old_word xor (old_word & clear_mask)))),
+            std::memory_order_relaxed);
         ++cur_word;
         shift_offset = bits_in_word - 1; // Word offset back to the beginning
     }
@@ -353,7 +355,7 @@ std::pair< uint64_t, uint32_t > Bitmap::next_dirty() noexcept {
             logical_off += (word_off * bits_in_word * _chunk_size); // Adjust for word
 
             // How long does the dirt stretch?
-            auto set_bit = __builtin_clzl(word);
+            auto set_bit = std::countl_zero(word);
             logical_off += set_bit * _chunk_size; // Adjust for bit within word
             // Consume as many consecutive set-bits as we can in the rest of the word
             while ((static_cast< int >(bits_in_word) > set_bit) && ((word >> (bits_in_word - (set_bit++) - 1)) & 0b1)) {
@@ -396,7 +398,7 @@ void Bitmap::dirty_region(uint64_t addr, uint64_t len) {
                                                      << (shift_offset - (bits_to_write - 1)));
             bits_left -= bits_to_write;
             auto old_word = cur_word->fetch_or(bits_to_set, std::memory_order_relaxed);
-            _dirty_chunks_est.fetch_add(__builtin_popcountll(old_word xor (old_word | bits_to_set)),
+            _dirty_chunks_est.fetch_add(std::popcount(old_word xor (old_word | bits_to_set)),
                                         std::memory_order_relaxed);
             ++cur_word;
             shift_offset = bits_in_word - 1; // Word offset back to the beginning
--- a/src/raid/raid1/raid1_resync_task.cpp
+++ b/src/raid/raid1/raid1_resync_task.cpp
@@ -262,7 +262,7 @@ void Raid1ResyncTask::__pause() noexcept {
 }
 
 // Abort any on-going resync task by moving to STOPPING and rejoin the thread
-uint32_t Raid1ResyncTask::stop() noexcept {
+void Raid1ResyncTask::stop() noexcept {
     auto lg = std::scoped_lock< std::mutex >(_launch_lock);
     // Targets SLEEPING or PAUSE → STOPPING (waits out ACTIVE first via RETRY_WITH_SLEEP).
     // Never CAS-es ACTIVE→STOPPING directly — this is what makes the ACTIVE→STOPPING
@@ -284,9 +284,6 @@ uint32_t Raid1ResyncTask::stop() noexcept {
         std::unreachable();
     });
     if (_resync_task.joinable()) _resync_task.join();
-    // I/O threads may still be between enqueue_write() and dequeue_write() at join time.
-    // Callers that care (e.g., ~Raid1DiskImpl) must drain all I/O before calling stop().
-    return static_cast< uint32_t >(_state_and_writes.count());
 }
 
 resync_state Raid1ResyncTask::__yield(std::chrono::microseconds const yield_for,
--- a/src/raid/raid1/raid1_resync_task.cpp
+++ b/src/raid/raid1/raid1_resync_task.cpp
@@ -264,7 +264,9 @@ void Raid1ResyncTask::__pause() noexcept {
 // Abort any on-going resync task by moving to STOPPING and rejoin the thread
 uint32_t Raid1ResyncTask::stop() noexcept {
     auto lg = std::scoped_lock< std::mutex >(_launch_lock);
-    // Terminate any ongoing resync task
+    // Targets SLEEPING or PAUSE → STOPPING (waits out ACTIVE first via RETRY_WITH_SLEEP).
+    // Never CAS-es ACTIVE→STOPPING directly — this is what makes the ACTIVE→STOPPING
+    // assert in __yield Phase 1 unreachable.
     __transition_to(resync_state::PAUSE, resync_state::STOPPING, [this](resync_state state) -> transition_result {
         switch (state) {
         case resync_state::IDLE: {
@@ -282,6 +284,8 @@ uint32_t Raid1ResyncTask::stop() noexcept {
         std::unreachable();
     });
     if (_resync_task.joinable()) _resync_task.join();
+    // I/O threads may still be between enqueue_write() and dequeue_write() at join time.
+    // Callers that care (e.g., ~Raid1DiskImpl) must drain all I/O before calling stop().
     return static_cast< uint32_t >(_state_and_writes.count());
 }
 
--- a/src/raid/raid1/raid1_resync_task.cpp
+++ b/src/raid/raid1/raid1_resync_task.cpp
@@ -309,9 +309,10 @@ resync_state Raid1ResyncTask::__yield(std::chrono::microseconds const yield_for,
         if (resync_state::STOPPING == cur_state) return cur_state;
 
         if (resync_state::PAUSE == cur_state) {
-            // I/O started during sleep - wait for it to complete
-            // Set to IDLE anticipating __resume() will transition PAUSE→ACTIVE
-            // This prevents busy-spinning and gives I/O threads priority
+            // A write is in flight; wait for dec_xchng_status_ifz to drive PAUSE→ACTIVE.
+            // Use IDLE as a sentinel so the next CAS always fails and reloads the real state —
+            // using PAUSE here would let __cas_state bypass the counter check and race with
+            // dec_xchng_status_ifz. Sleep to yield bandwidth to the write path.
             cur_state = resync_state::IDLE;
             std::this_thread::sleep_for(yield_for);
         }
--- a/src/raid/raid1/raid1_resync_task.cpp
+++ b/src/raid/raid1/raid1_resync_task.cpp
@@ -17,8 +17,6 @@ Raid1ResyncTask::Raid1ResyncTask(std::shared_ptr< raid1::Bitmap >& bitmap, uint6
         _io_size(io_size),
         _max_size(max_io),
         _offset(offset),
-        _resync_state(static_cast< uint8_t >(resync_state::IDLE)), // Initial store, can't use helper
-        _outstanding_writes(0U),
         _resync_task() {
     if (!_dirty_bitmap) throw std::runtime_error("No Bitmap");
 }
@@ -30,7 +28,7 @@ Raid1ResyncTask::~Raid1ResyncTask() noexcept {
 template < typename StateHandler >
 bool Raid1ResyncTask::__transition_to(resync_state initial, resync_state target, StateHandler&& handler) noexcept {
     auto cur_state = initial;
-    while (!__cas_state_weak(cur_state, target)) {
+    while (!__cas_state(cur_state, target)) {
         auto [next_state, action] = handler(cur_state);
 
         switch (action) {
@@ -55,8 +53,7 @@ void Raid1ResyncTask::_start(std::string str_uuid, std::shared_ptr< MirrorDevice
     RLOGD("Resync Task created for [uuid:{}]", str_uuid)
     // Wait to become Available & IDLE
     auto cur_state = resync_state::IDLE;
-    while (dirty_mirror->unavail.test(std::memory_order_acquire) ||
-           !__cas_state_weak(cur_state, resync_state::ACTIVE)) {
+    while (dirty_mirror->unavail.test(std::memory_order_acquire) || !__cas_state(cur_state, resync_state::ACTIVE)) {
         // If we're stopped or another we should exit
         if (resync_state::STOPPING == cur_state) break;
 
@@ -110,7 +107,7 @@ void Raid1ResyncTask::_start(std::string str_uuid, std::shared_ptr< MirrorDevice
     // If stopped, end now.
     if (resync_state::STOPPING == cur_state) {
         RLOGI("Resync Task Stopped for [uuid:{}] to: {}", str_uuid, *dirty_mirror->disk)
-        __cas_state_strong(cur_state, resync_state::IDLE);
+        _state_and_writes.xchng_status(cur_state, resync_state::IDLE);
         return;
     }
 
@@ -121,7 +118,7 @@ void Raid1ResyncTask::_start(std::string str_uuid, std::shared_ptr< MirrorDevice
     RLOGD("Resync Task Finished for [uuid:{}] to: {}", str_uuid, *dirty_mirror->disk)
 
     // Open up I/O Again
-    __cas_state_strong(cur_state, resync_state::IDLE);
+    _state_and_writes.xchng_status(cur_state, resync_state::IDLE);
 }
 
 void Raid1ResyncTask::launch(std::string const& str_uuid, std::shared_ptr< MirrorDevice > clean_mirror,
@@ -285,7 +282,7 @@ uint32_t Raid1ResyncTask::stop() noexcept {
         std::unreachable();
     });
     if (_resync_task.joinable()) _resync_task.join();
-    return _outstanding_writes.load(std::memory_order_acquire);
+    return static_cast< uint32_t >(_state_and_writes.count());
 }
 
 resync_state Raid1ResyncTask::__yield(std::chrono::microseconds const yield_for,
@@ -293,7 +290,7 @@ resync_state Raid1ResyncTask::__yield(std::chrono::microseconds const yield_for,
     auto cur_state = resync_state::ACTIVE;
 
     // Phase 1: Transition ACTIVE→SLEEPING (give I/O a chance to interrupt)
-    while (!__cas_state_weak(cur_state, resync_state::SLEEPING)) {
+    while (!__cas_state(cur_state, resync_state::SLEEPING)) {
         if (resync_state::STOPPING == cur_state) return cur_state;
     }
     cur_state = resync_state::SLEEPING;
@@ -306,7 +303,7 @@ resync_state Raid1ResyncTask::__yield(std::chrono::microseconds const yield_for,
     }
 
     // Phase 3: Transition SLEEPING→ACTIVE (resume resync)
-    while (!__cas_state_weak(cur_state, resync_state::ACTIVE)) {
+    while (!__cas_state(cur_state, resync_state::ACTIVE)) {
         if (resync_state::STOPPING == cur_state) return cur_state;
 
         if (resync_state::PAUSE == cur_state) {
--- a/src/raid/raid1/raid1_resync_task.cpp
+++ b/src/raid/raid1/raid1_resync_task.cpp
@@ -291,7 +291,9 @@ resync_state Raid1ResyncTask::__yield(std::chrono::microseconds const yield_for,
 
     // Phase 1: Transition ACTIVE→SLEEPING (give I/O a chance to interrupt)
     while (!__cas_state(cur_state, resync_state::SLEEPING)) {
-        if (resync_state::STOPPING == cur_state) return cur_state;
+        // STOPPING here is unreachable: stop() only CAS SLEEPING→STOPPING or PAUSE→STOPPING,
+        // never ACTIVE→STOPPING, so this loop exits on the first successful CAS.
+        DEBUG_ASSERT_NE(cur_state, resync_state::STOPPING, "impossible ACTIVE→STOPPING transition"); // LCOV_EXCL_LINE
     }
     cur_state = resync_state::SLEEPING;
 
--- a/src/raid/raid1/raid1_resync_task.hpp
+++ b/src/raid/raid1/raid1_resync_task.hpp
@@ -106,7 +106,7 @@ public:
                 std::shared_ptr< MirrorDevice > dirty_mirror, std::function< void() >&& complete);
 
     // Generic method to move Resync StateMachine to STOPPING
-    uint32_t stop() noexcept; // Returns the _outstanding_writes cnt here
+    uint32_t stop() noexcept; // Returns outstanding write count at time of stop
 
     inline void enqueue_write() noexcept {
         _state_and_writes.set_atomic_value([](auto& cnt, auto& /*status*/) {
--- a/src/raid/raid1/raid1_resync_task.hpp
+++ b/src/raid/raid1/raid1_resync_task.hpp
@@ -3,6 +3,9 @@
 #include <atomic>
 #include <chrono>
 #include <thread>
+#include <functional>
+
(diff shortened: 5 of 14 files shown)


Ilyin-Ilya (commented):
Multi-agent review of the `atomic_status_counter` refactor. The core fix is sound and well-reasoned; findings below are ordered by severity. One confirmed false positive was filtered out (the `cnt++`→`__pause()` window concern in `enqueue_write` — safe because the actual write I/O is submitted only after `enqueue_write()` returns in full, so resync running inside that window has no data to race against).

---

## Warnings

**[WARNING] `stop()` return value may be stale or non-zero**
`stop()` joins the resync thread and then returns `_state_and_writes.count()`. If any I/O thread is still between `enqueue_write()` and its subsequent disk write at join time, the returned count will be non-zero even though the caller may expect zero. The old code had the same shape, but the PR description promotes this value as meaningful. Either document that a non-zero return is normal and callers must not treat it as an error, or drain the counter before returning.

**[WARNING] `__yield` Phase 1 STOPPING check: confirm unreachability**
The STOPPING branch was removed from Phase 1 of `__yield` and replaced with `DEBUG_ASSERT_NE`. The comment says `stop()` never issues ACTIVE→STOPPING. This is correct only if `stop()` exclusively targets SLEEPING and PAUSE states for the STOPPING CAS. Please add a brief inline comment at the `stop()` CAS site naming the exact states it targets, so reviewers can verify the assertion cannot fire in release builds without auditing the full state machine.

**[WARNING] Memory ordering of `sisl::atomic_status_counter` operations unverified**
The old code used explicit `memory_order_release` / `memory_order_acquire` on the separate `_resync_state` and `_outstanding_writes` atomics. The new `set_atomic_value` / `dec_xchng_status_ifz` calls wrap those into sisl internals whose ordering is not visible in this diff. If sisl uses `memory_order_relaxed` internally for the CAS, the happens-before edge that ensures resync sees writes completed before PAUSE→ACTIVE is lost. Please verify (or add a comment citing) the memory order used by `dec_xchng_status_ifz`, and add a `static_assert(decltype(_state_and_writes)::is_always_lock_free)` to confirm the 64-bit CAS assumption — the PR description mentions "5-byte packed" in one place and "64-bit CAS word" in another; these are contradictory and worth resolving explicitly.

**[WARNING] `ublkpp` shared library missing `-latomic` in `src/CMakeLists.txt`**
`-latomic` is added to `raid1` and `example/CMakeLists.txt`, but the top-level `ublkpp` shared library target in `src/CMakeLists.txt` does not explicitly link it. The Conan `system_libs` entry propagates to consumers via package info, but does not add `atomic` to the `.so`'s own `DT_NEEDED` section. Consumers that link the raw CMake target (without Conan) may get runtime linker failures. Suggest adding `$<$<PLATFORM_ID:Linux>:atomic>` to the `ublkpp` shared lib target the same way it was added to `raid1`.

**[WARNING] C++23 `_min_cppstd` bump is a breaking consumer change — version should be minor, not patch**
Raising `_min_cppstd` from 20 to 23 breaks any consumer whose toolchain does not yet support C++23. CHANGELOG lists this as `0.22.1` (patch), but a mandatory minimum standard bump is a breaking API/ABI constraint and warrants at least a minor bump (`0.23.0`) with a bolded **Breaking** note in the changelog, consistent with how `0.22.0` documented its breaking `open_for_uring` signature change.

---

## Test Coverage Gaps

**[WARNING] Race 1 window is too narrow for probabilistic detection; document this in the test source**
The PR description correctly notes that the ~2–5 instruction Race 1 window cannot be reliably triggered and that correctness rests on an analytical argument. That caveat exists only in the PR description, not in the test file. Add a comment in `Raid1Concurrency.EnqueueDequeueRace` noting that Race 1 is verified analytically (the test primarily exercises Race 2 under sanitizer overhead) so future maintainers understand why the test does not assert Race 1 deterministically.

**[WARNING] Race 2 (`__pause()` spin) is invisible without ASan**
`writes_in_flight` in the test is incremented after `enqueue_write()` returns, so a plain Debug build cannot observe a write racing the resync spin. The test plan bullet notes "ASan is required for reliable detection" — mirror this requirement in the test source with a comment (or a `GTEST_SKIP` when sanitizers are absent) so CI configurations that run only the Debug suite do not get false confidence.

**[NIT] `sleep_for(10ms)` to await resync start is fragile under TSan / heavy load**
The fixed sleep to wait for the resync thread to reach its spin loop can cause spurious failures on loaded CI runners or under ThreadSanitizer overhead. A small semaphore or `std::atomic` sentinel written by the mock before it begins the slow copy would make the synchronization deterministic.

---

## Pre-existing Nits (not introduced by this PR)

**[NIT] Rollback CAS failure in `__swap_device` is silently dropped**
If the rollback `compare_exchange_strong(new_read_route, cur_route)` fails, the failure is ignored. This is pre-existing behavior. A `RLOGD` on failure would aid debugging.

**[NIT] `auto word = 0UL` in `bitmap.cpp::next_dirty` is a portability hazard**
`std::countl_zero(word)` operates on `unsigned long`, which is 32 bits on LLP64 (Windows). The bitmap word type should be `uint64_t` explicitly. This PR does not regress the old `__builtin_clzl` behavior, but the migration to `<bit>` is a good opportunity to fix the underlying type.

---

*Reviewed by 4 parallel analysis agents (correctness, test coverage, build system/C++23, refactor quality) with cross-check for false positives. The `enqueue_write` non-atomicity finding raised by one agent was determined to be a false positive and is not reported here.*

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.22.1
+- raid1: Fix dequeue/resume race — `_resync_state` and `_outstanding_writes` are now packed into a single `sisl::atomic_status_counter` so the counter decrement and PAUSE→ACTIVE transition are one indivisible CAS; `__resume()` is removed (SDSTOR-21927)
+- raid1: Fix enqueue/pause race — `enqueue_write()` now always calls `__pause()` on every enqueue, not only the first; previously a concurrent second enqueuer could skip `__pause()` while the first was still establishing it, allowing resync to overwrite an in-flight write with stale data
+- raid1: Replace GCC `__builtin_popcount`/`__builtin_clz`/`__builtin_ctz` with C++23 `std::popcount`/`std::countl_zero`/`std::countr_zero`
+- build: `libatomic` is now declared as a Conan system lib on Linux — propagated automatically to consumers, no downstream changes required
+
 ## 0.22.0
 - raid1: Fix multi-queue idle probe race conditions — probes now start only when all queues are idle, mutex serializes concurrent launch/stop calls, `open_for_uring` counts queue threads for accurate `nr_hw_queues`
```
Ilyin-Ilya, line 14: **[WARNING]** The C++23 minimum standard bump (`_min_cppstd` 20→23) is a breaking consumer change — any downstream toolchain still on C++20 will get a Conan validation error. This entry should document it explicitly:
```
- build: minimum C++ standard raised from C++20 to C++23
```
Also consider whether this warrants a minor version bump (`0.23.0`) rather than a patch, consistent with how `0.22.0` called out its breaking `open_for_uring` signature change.

conanfile.py
```diff
@@ -47,7 +47,7 @@ class UBlkPPConan(ConanFile):
                         )
 
     def _min_cppstd(self):
-        return 20
+        return 23
```
Ilyin-Ilya, line 51: **[WARNING]** Raising `_min_cppstd` from 20 to 23 is a breaking change for consumers still on C++20. The CHANGELOG entry for `0.22.1` does not mention this. Please add a **Breaking** bullet and consider bumping to `0.23.0` to match the project's own precedent (see `0.22.0`'s breaking note for `open_for_uring`).

src/raid/raid1/raid1_resync_task.hpp
```diff
@@ -54,34 +57,28 @@ class Raid1ResyncTask {
...
 
-    std::atomic_uint8_t _resync_state;
-    std::atomic_uint32_t _outstanding_writes;
+    // Packs resync_state (status) + outstanding-write count (int32_t counter) into one 64-bit
+    // word operated on with a single CAS. Key guarantee: dec_xchng_status_ifz() decrements the
+    // counter and, only if it reaches zero AND status==PAUSE, atomically transitions to ACTIVE —
+    // eliminating the window where a concurrent enqueue_write() could see count==0 while state
+    // remains PAUSE, early-exit __pause(), and then race with a __resume() that fires afterward.
+    sisl::atomic_status_counter< resync_state, resync_state::IDLE > _state_and_writes;
     std::mutex _launch_lock;
```
Ilyin-Ilya, line 80: **[WARNING]** The code comment here says "one 64-bit word operated on with a single CAS" but that is inaccurate. `sisl::atomic_status_counter` stores a `std::atomic<_status_counter>` where `_status_counter` is a **5-byte struct**:

- `int32_t counter` → 4 bytes
- `uint8_t status` → 1 byte
- `#pragma pack(1)` removes padding → **5 bytes total**

x86-64 has native atomic instructions only for 1, 2, 4, and 8 bytes — not 5 — so `std::atomic<_status_counter>` cannot use a single `CMPXCHG` and GCC falls back to libatomic. `is_always_lock_free` is **false** for this type. The `uint64_t` conversion operator in sisl exists for inspection only, not for the atomic ops.

This is correct for correctness — libatomic serializes via address-indexed spinlocks and uses `acq_rel` ordering — and the libatomic dependency added in `CMakeLists.txt` / `conanfile.py` is legitimate and necessary.

Two asks:
1. **Fix the comment** from "one 64-bit CAS word" to something like: `// 5-byte struct (int32_t counter + uint8_t status, #pragma pack(1)); not lock-free — serialized by libatomic spinlocks.`
2. **Note the performance implication**: every `enqueue_write()`, `dequeue_write()`, and `__cas_state()` call goes through a libatomic spinlock. Almost certainly fine here (write frequency << resync frequency), but worth documenting so future readers do not assume hardware-CAS throughput.

Ilyin-Ilya (commented):
Inline review (supersedes the top-level comment above). Reviewed by 4 parallel agents across correctness, test coverage, build system/C++23, and refactor quality. One false positive was filtered: the `cnt++`→`__pause()` window in `enqueue_write` is safe because the actual disk write is submitted only after `enqueue_write()` returns.

src/raid/raid1/raid1_resync_task.hpp
```diff
@@ -113,18 +106,28 @@ class Raid1ResyncTask {
...
+    uint32_t stop() noexcept; // Returns outstanding write count at time of stop
 
     inline void enqueue_write() noexcept {
-        auto const old_val = _outstanding_writes.fetch_add(1, std::memory_order_release);
-        if (0 == old_val) __pause();
-        DEBUG_ASSERT_LT(old_val, UINT32_MAX, "Outstanding Write Count Overflowed!");
+        _state_and_writes.set_atomic_value([](auto& cnt, auto& /*status*/) {
+            ++cnt;
+            return true;
+        });
```
Ilyin-Ilya, line 131: **[WARNING]** The increment and the `__pause()` call are two separate steps with a window between them. This is safe because the caller only submits the actual disk write *after* `enqueue_write()` returns (including after `__pause()` completes). That ordering contract is stated in the test but not here — a brief comment would prevent future readers from raising the same concern:
```cpp
// Increment first, then establish pause. Safe because the caller submits I/O
// only after this function returns, so no data is on disk during this window.
```

src/raid/raid1/raid1_resync_task.hpp
```diff
@@ -113,18 +106,28 @@ class Raid1ResyncTask {
...
-        if (1 == old_val) __resume();
-        DEBUG_ASSERT_GT(old_val, 0, "Outstanding Write Count Underflowed!");
+        DEBUG_ASSERT_GT(_state_and_writes.count(), 0, "Outstanding Write Count Underflowed!");
+        // Atomically decrement; if counter hits zero while state is PAUSE, transition to ACTIVE.
+        // This single CAS closes the race where a concurrent enqueue could see old_val==0,
+        // find state still PAUSE (from the prior write), and early-exit __pause() — only for
+        // __resume() to then clear PAUSE before the new write's I/O is submitted.
+        _state_and_writes.dec_xchng_status_ifz(resync_state::PAUSE, resync_state::ACTIVE);
     }
 };
```
Ilyin-Ilya, line 150: **[NIT]** The `DEBUG_ASSERT_GT` check fires before the decrement, but in a Release build it is a no-op. If `dequeue_write()` is called more times than `enqueue_write()` (e.g., a bug in a caller), the `int32_t` counter goes negative in Release and `dec_xchng_status_ifz(PAUSE, ACTIVE)` would not trigger correctly (counter never hits 0). Worth a note in the comment, or a cheap release-visible check if underflow is a plausible bug surface.

src/raid/raid1/raid1_resync_task.cpp
```diff
@@ -285,16 +282,18 @@ uint32_t Raid1ResyncTask::stop() noexcept {
         std::unreachable();
     });
     if (_resync_task.joinable()) _resync_task.join();
```
Ilyin-Ilya, line 286: **[WARNING]** `stop()` joins the resync task and then reads the counter. If I/O threads are still between `enqueue_write()` and `dequeue_write()` at join time, the returned count can be non-zero — which may surprise callers. The old code had the same shape, but the PR description highlights this return value as meaningful. Please document the contract: is a non-zero return here expected and benign, or does the caller guarantee I/O is drained before calling `stop()`?

src/raid/raid1/raid1_resync_task.cpp
```diff
@@ -285,16 +282,18 @@ uint32_t Raid1ResyncTask::stop() noexcept {
...
 
 resync_state Raid1ResyncTask::__yield(std::chrono::microseconds const yield_for,
                                       std::chrono::microseconds const spin_time) noexcept {
     auto cur_state = resync_state::ACTIVE;
 
     // Phase 1: Transition ACTIVE→SLEEPING (give I/O a chance to interrupt)
-    while (!__cas_state_weak(cur_state, resync_state::SLEEPING)) {
-        if (resync_state::STOPPING == cur_state) return cur_state;
+    while (!__cas_state(cur_state, resync_state::SLEEPING)) {
+        // STOPPING here is unreachable: stop() only CAS SLEEPING→STOPPING or PAUSE→STOPPING,
```
Ilyin-Ilya, line 295: **[WARNING]** The `DEBUG_ASSERT_NE` relies on `stop()` never CAS-ing ACTIVE→STOPPING — only SLEEPING→STOPPING or PAUSE→STOPPING. This is correct per the current `stop()` handler, but it is a non-obvious invariant with no release-build protection. Please add a brief comment at the CAS site inside `stop()` listing the exact source states it targets (SLEEPING, PAUSE) so a future reader can verify the assert is unreachable without tracing the full state machine.

src/raid/raid1/tests/concurrency/concurrent_enqueue_dequeue.cpp
```diff
@@ -0,0 +1,125 @@
...
+//
+// Ordering contract:
+//   enqueue_write()         — pause is established; resync may not run past this point
+//   writes_in_flight.fetch_add  — record that we are in-flight
+//   [write runs]
+//   writes_in_flight.fetch_sub  — mark flight complete
+//   dequeue_write()         — release the pause
+//
+// This bracket guarantees: if the mock write fires while writes_in_flight > 0, the
+// invariant was violated. The ordering avoids false positives (no inflation before
```
Ilyin-Ilya, line 44: **[WARNING]** This line says TSAN is the authoritative signal, but the test source doesn't explain *why* — specifically:
- **Race 1** (the 2–5 instruction window in the old `fetch_sub` / `__resume()` pair) cannot be triggered probabilistically. Its fix is analytically correct, not test-verified here.
- **Race 2** (`__pause()` spin) is detectable by this test, but only reliably under ASan/TSan where instrumentation widens the spin window; in a plain Debug build `writes_in_flight` is incremented *after* `enqueue_write()` returns, so an overlap during the spin is invisible.

Please add a short note at the top of the test (before `TEST(...)`) clarifying these limits so CI configurations that skip sanitizers don't treat a green run as a regression guard.

src/raid/raid1/tests/concurrency/concurrent_enqueue_dequeue.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    auto uuid = boost::uuids::string_generator()(test_uuid);
+    auto mirror_a = std::make_shared< MirrorDevice >(uuid, device_a);
+    auto mirror_b = std::make_shared< MirrorDevice >(uuid, device_b);
+
+    auto superbitmap_buf = make_test_superbitmap();
+    auto bitmap = std::make_shared< Bitmap >(Gi, 32 * Ki, 4 * Ki, superbitmap_buf.get());
+    bitmap->dirty_region(0, Gi);
+
+    constexpr uint32_t io_size = 512 * Ki;
```
Ilyin-Ilya, line 98: **[NIT]** The fixed 10ms sleep to wait for the resync thread to begin copying is fragile under heavy load or TSan instrumentation overhead. If the resync hasn't reached its copy loop when the I/O threads start, the test exercises only the IDLE→ACTIVE path, which is not the interesting case. Consider replacing with a sentinel:
```cpp
std::atomic<bool> resync_started{false};
// In device_b WRITE mock, before the sleep:
resync_started.store(true, std::memory_order_release);
// Then:
while (!resync_started.load(std::memory_order_acquire)) std::this_thread::yield();
```

src/raid/raid1/bitmap.cpp
```diff
@@ -353,7 +355,7 @@ std::pair< uint64_t, uint32_t > Bitmap::next_dirty() noexcept {
             logical_off += (word_off * bits_in_word * _chunk_size); // Adjust for word
 
             // How long does the dirt stretch?
-            auto set_bit = __builtin_clzl(word);
+            auto set_bit = std::countl_zero(word);
             logical_off += set_bit * _chunk_size; // Adjust for bit within word
```
Ilyin-Ilya, line 359: **[NIT]** `word` is declared as `auto word = 0UL` (from the surrounding context), so `std::countl_zero(word)` instantiates on `unsigned long` — 32 bits on LLP64 (Windows). The bitmap word type is `uint64_t`, so this is a latent portability hazard. `__builtin_clzl` had the same problem, so this PR doesn't introduce a regression, but the `<bit>` migration is a good opportunity to fix it:
```cpp
uint64_t word = 0;
```

CHANGELOG.md
Ilyin-Ilya, line 14: **[WARNING]** The C++23 minimum standard bump (`_min_cppstd` 20→23) is a breaking consumer change — any downstream toolchain still on C++20 will get a Conan validation error. This entry should document it explicitly:
```
- build: minimum C++ standard raised from C++20 to C++23
```
Also consider whether this warrants a minor version bump (`0.23.0`) rather than a patch, consistent with how `0.22.0` called out its breaking `open_for_uring` signature change.

conanfile.py
Ilyin-Ilya, line 51: **[WARNING]** Raising `_min_cppstd` from 20 to 23 is a breaking change for consumers still on C++20. The CHANGELOG entry for `0.22.1` does not mention this. Please add a **Breaking** bullet and consider bumping to `0.23.0` to match the project's own precedent (see `0.22.0`'s breaking note for `open_for_uring`).

src/raid/raid1/raid1_resync_task.hpp
Ilyin-Ilya, line 80: **[WARNING]** Two issues here:

1. The PR description says "5-byte packed" in the build section but this comment says "64-bit CAS word". These are contradictory — a 5-byte struct is not natively atomic on x86-64 and would require libatomic via software locking, while an 8-byte struct uses a single `CMPXCHG8B`. Please clarify which it actually is and add:
```cpp
static_assert(decltype(_state_and_writes)::is_always_lock_free,
    "atomic_status_counter must be lock-free on this platform");
```
If it is *not* always lock-free, the locking is silently hidden inside libatomic — that's an accepted trade-off, but should be documented.

2. The memory ordering of `set_atomic_value` and `dec_xchng_status_ifz` is not visible in this diff. The old code had explicit `memory_order_release` (on `fetch_add`) and `memory_order_acquire` (on `fetch_sub`) to establish the happens-before between write completion and PAUSE→ACTIVE. Please add a comment or verify that `sisl::atomic_status_counter` uses at least `acq_rel` for its CAS, otherwise the happens-before edge is lost.

src/raid/raid1/raid1_resync_task.hpp
Ilyin-Ilya, line 131: **[WARNING]** The increment and the `__pause()` call are two separate steps with a window between them. This is safe because the caller only submits the actual disk write *after* `enqueue_write()` returns (including after `__pause()` completes). That ordering contract is stated in the test but not here — a brief comment would prevent future readers from raising the same concern:
```cpp
// Increment first, then establish pause. Safe because the caller submits I/O
// only after this function returns, so no data is on disk during this window.
```

src/raid/raid1/raid1_resync_task.hpp
Ilyin-Ilya, line 150: **[NIT]** The `DEBUG_ASSERT_GT` check fires before the decrement, but in a Release build it is a no-op. If `dequeue_write()` is called more times than `enqueue_write()` (e.g., a bug in a caller), the `int32_t` counter goes negative in Release and `dec_xchng_status_ifz(PAUSE, ACTIVE)` would not trigger correctly (counter never hits 0). Worth a note in the comment, or a cheap release-visible check if underflow is a plausible bug surface.

src/raid/raid1/raid1_resync_task.cpp
Ilyin-Ilya, line 286: **[WARNING]** `stop()` joins the resync task and then reads the counter. If I/O threads are still between `enqueue_write()` and `dequeue_write()` at join time, the returned count can be non-zero — which may surprise callers. The old code had the same shape, but the PR description highlights this return value as meaningful. Please document the contract: is a non-zero return here expected and benign, or does the caller guarantee I/O is drained before calling `stop()`?

src/raid/raid1/raid1_resync_task.cpp
Ilyin-Ilya, line 295: **[WARNING]** The `DEBUG_ASSERT_NE` relies on `stop()` never CAS-ing ACTIVE→STOPPING — only SLEEPING→STOPPING or PAUSE→STOPPING. This is correct per the current `stop()` handler, but it is a non-obvious invariant with no release-build protection. Please add a brief comment at the CAS site inside `stop()` listing the exact source states it targets (SLEEPING, PAUSE) so a future reader can verify the assert is unreachable without tracing the full state machine.

src/raid/raid1/tests/concurrency/concurrent_enqueue_dequeue.cpp
Ilyin-Ilya, line 44: **[WARNING]** This line says TSAN is the authoritative signal, but the test source doesn't explain *why* — specifically:
- **Race 1** (the 2–5 instruction window in the old `fetch_sub` / `__resume()` pair) cannot be triggered probabilistically. Its fix is analytically correct, not test-verified here.
- **Race 2** (`__pause()` spin) is detectable by this test, but only reliably under ASan/TSan where instrumentation widens the spin window; in a plain Debug build `writes_in_flight` is incremented *after* `enqueue_write()` returns, so an overlap during the spin is invisible.

Please add a short note at the top of the test (before `TEST(...)`) clarifying these limits so CI configurations that skip sanitizers don't treat a green run as a regression guard.

src/raid/raid1/tests/concurrency/concurrent_enqueue_dequeue.cpp
Ilyin-Ilya, line 98: **[NIT]** The fixed 10ms sleep to wait for the resync thread to begin copying is fragile under heavy load or TSan instrumentation overhead. If the resync hasn't reached its copy loop when the I/O threads start, the test exercises only the IDLE→ACTIVE path, which is not the interesting case. Consider replacing with a sentinel:
```cpp
std::atomic<bool> resync_started{false};
// In device_b WRITE mock, before the sleep:
resync_started.store(true, std::memory_order_release);
// Then:
while (!resync_started.load(std::memory_order_acquire)) std::this_thread::yield();
```

src/raid/raid1/bitmap.cpp
Ilyin-Ilya, line 359: **[NIT]** `word` is declared as `auto word = 0UL` (from the surrounding context), so `std::countl_zero(word)` instantiates on `unsigned long` — 32 bits on LLP64 (Windows). The bitmap word type is `uint64_t`, so this is a latent portability hazard. `__builtin_clzl` had the same problem, so this PR doesn't introduce a regression, but the `<bit>` migration is a good opportunity to fix it:
```cpp
uint64_t word = 0;
```

Ilyin-Ilya (commented):
Inline review (supersedes the top-level comment above). Reviewed by 4 parallel agents across correctness, test coverage, build system/C++23, and refactor quality. One false positive was filtered: the `cnt++`→`__pause()` window in `enqueue_write` is safe because the actual disk write is submitted only after `enqueue_write()` returns.

Ilyin-Ilya (commented):
test

Ilyin-Ilya (commented):
test

src/raid/raid1/bitmap.cpp
```diff
@@ -353,7 +355,7 @@ std::pair< uint64_t, uint32_t > Bitmap::next_dirty() noexcept {
             logical_off += (word_off * bits_in_word * _chunk_size); // Adjust for word
 
             // How long does the dirt stretch?
-            auto set_bit = __builtin_clzl(word);
+            auto set_bit = std::countl_zero(word);
```
Ilyin-Ilya, line 358: **[NIT]** `word` is declared as `auto word = 0UL`, so `std::countl_zero(word)` instantiates on `unsigned long` — 32 bits on LLP64. The bitmap word type is `uint64_t`; using `uint64_t word = 0` here would make the type unambiguous and portable.

src/raid/raid1/raid1_resync_task.hpp
szmyd: Just extend the status to 32bits right? So it's <32bit><32bit> packed struct. No performance implication

CHANGELOG.md
szmyd: Nope, it's being fixed here; was broken before.

conanfile.py
szmyd: see above.

CHANGELOG.md
szmyd: `std::expected` is C++23 and we use it throughout the codebase, so `_min_cppstd = 20` was already wrong — any consumer on C++20 would fail at compile time rather than at Conan validation. This PR makes the formal check match the actual requirement; it does not introduce a new incompatibility. No version bump needed.

conanfile.py
szmyd: `std::expected` is C++23 and we use it throughout the codebase, so `_min_cppstd = 20` was already wrong — any consumer on C++20 would fail at compile time rather than at Conan validation. This PR makes the formal check match the actual requirement; it does not introduce a new incompatibility. No version bump needed.

src/raid/raid1/raid1_resync_task.hpp
```diff
@@ -113,18 +118,32 @@ class Raid1ResyncTask {
                 std::shared_ptr< MirrorDevice > dirty_mirror, std::function< void() >&& complete);
 
     // Generic method to move Resync StateMachine to STOPPING
-    uint32_t stop() noexcept; // Returns the _outstanding_writes cnt here
+    void stop() noexcept; // Returns outstanding write count at time of stop
```
Ilyin-Ilya: **[NIT]** The comment on `stop()` says "Returns outstanding write count at time of stop" but the return type is now `void`. The comment is stale — it should either be removed or replaced with the actual contract:
```cpp
void stop() noexcept; // Callers must drain I/O before calling; outstanding count at join time is not a meaningful signal.
```

src/raid/raid1/raid1_resync_task.hpp
```diff
@@ -54,34 +57,40 @@ class Raid1ResyncTask {
...
+    // status==PAUSE, atomically transitions to ACTIVE — eliminating the window where a concurrent
+    // enqueue_write() could see count==0 while state remains PAUSE, early-exit __pause(), and
+    // then race with a __resume() that fires afterward.
+    sisl::atomic_status_counter< resync_state, resync_state::IDLE > _state_and_writes;
+    // Verify that resync_state is still 32-bit so the combined struct stays lock-free.
+    static_assert(
+        [] {
+            struct s {
+                int32_t c;
+                resync_state st;
```
Ilyin-Ilya, line 74: **[NIT]** The `static_assert` checks lock-free-ness of a local proxy `struct s { int32_t c; resync_state st; }`, not the actual `sisl::_status_counter` type that `_state_and_writes` wraps. If sisl adds `#pragma pack`, extra fields, or a different layout, the proxy can return `is_always_lock_free == true` while the real type is not. The assert would silently stop guarding the intended invariant.

Consider asserting on the actual field type directly:
```cpp
static_assert(decltype(_state_and_writes)::is_always_lock_free,
    "_state_and_writes must be lock-free on this platform");
```
If `sisl::atomic_status_counter` doesn't expose `is_always_lock_free`, a note documenting why the proxy is an adequate proxy would also close this.

src/raid/raid1/tests/concurrency/concurrent_enqueue_dequeue.cpp
```diff
@@ -0,0 +1,135 @@
+#include "test_raid1_common.hpp"
+
+#include <atomic>
+#include <barrier>
```
Ilyin-Ilya, line 4: **[NIT]** `#include <barrier>` is unused — the test replaced the intended `std::barrier` synchronisation with an atomic sentinel (`resync_started`). Remove the include to keep the file clean and avoid a compiler warning with `-Wunused-includes`.

conanfile.py
```diff
@@ -146,6 +146,8 @@ def package(self):
 
     def package_info(self):
         self.cpp_info.requires = ["sisl::cache", "isa-l::isa-l", "ublksrv::ublksrv"]
```
Ilyin-Ilya, line 148: **[WARNING]** The PR adds `-latomic` to `raid1` (OBJECT library), `test_raid1`, `example`, and `fio_engine`, but not to the main `ublkpp` *shared* library target in `src/CMakeLists.txt`. Because `raid1` is an OBJECT library linked via `$<TARGET_OBJECTS:raid1>`, its link requirements do not propagate to the shared library's own link command.

`conanfile.py` propagates `system_libs = ["atomic"]` to Conan consumers, so installed-package users are covered. However, the *build-tree* shared library (`libublkpp.so`) may fail to link in configurations where the linker requires an explicit `-latomic` (e.g., GCC on ARM/32-bit hosts, or when using `-static-libstdc++`). Consider adding:

```cmake
# src/CMakeLists.txt, in the ublkpp target_link_libraries:
$<$<PLATFORM_ID:Linux>:atomic>
```

to close the gap between the Conan-installed path and the raw-CMake build path.

Ilyin-Ilya:
## Review summary

The core fix is correct — replacing two separate atomics with `sisl::atomic_status_counter` genuinely closes both races, and the PR description's analysis is accurate. Brian's responses on `_min_cppstd` (already a C++23 requirement) and the 5-byte struct (upgrading status to 32-bit in sisl) are both accepted. Four additional findings below; nothing is a blocker.

---

### [WARNING] `-latomic` missing from the `ublkpp` shared library target (`src/CMakeLists.txt`)

`raid1` is an OBJECT library. Its `target_link_libraries($<PLATFORM_ID:Linux>:atomic>)` does not propagate through `$<TARGET_OBJECTS:raid1>` to `libublkpp.so`. Conan consumers are covered via `system_libs = ["atomic"]`, but a raw-CMake build of the shared library on ARM/32-bit or with `-static-libstdc++` may fail to link. Inline comment on `conanfile.py`.

### [NIT] Stale comment on `stop()` declaration in `raid1_resync_task.hpp`

`void stop() noexcept; // Returns outstanding write count at time of stop` — return type is `void`, comment still says "Returns...". Should be removed or updated. Inline comment at the declaration.

### [NIT] `static_assert` guards a proxy struct, not the actual `sisl::atomic_status_counter` type

The assert checks `std::atomic<struct s>::is_always_lock_free` on a hand-rolled `{int32_t c; resync_state st;}` rather than `decltype(_state_and_writes)::is_always_lock_free`. If sisl's internal layout diverges, the assert silently stops tracking the real type. Inline comment at the `static_assert`.

### [NIT] `#include <barrier>` is unused in the regression test

The test switched from a `std::barrier` design to an atomic sentinel; the include was not cleaned up. Inline comment on the test file.

Ilyin-Ilya (approved):

PR #209, opened 2026-04-28
test(raid1): add regression for IDLE→STOPPING race after resync compl…

### LINKED ISSUE #205: raid1: Resync hangs in active+dirty+syncing after partition move when write races resync completion
## Summary

After a partition move (new replica added), the RAID1 resync gets stuck in `active+dirty+syncing` indefinitely. The `ublk_dirty_pages` metric reports 0 but the state never transitions to clean.

Surfaced by: (`partition-move-2-0-nvme cold_with_io` test timing out at 12000s).

## Symptom

```
[13:30:50] __become_clean  → superblock written (age:2)   ← first resync completes
[13:30:57] Partition change: sdb → sdd                    ← new replica added
[13:30:58] Bitmap cleared on sdd, superblock written (age:18)
[13:30:58] State: active+dirty+syncing                    ← second resync starts
...
[16:52:29] State: active+dirty+syncing                    ← still stuck 3+ hours later
[16:52:29] Timeout (12000s elapsed)

$ curl localhost:5000/metrics | grep dirty
ublk_dirty_pages{...} 0                                   ← metric shows 0 but never became clean
```

## What Is Known

- The second resync is launched correctly after the swap (`toggle_resync(true)` fires, route=DEVA, `new_device=true` so `dirty_region(0, capacity())` marks all pages dirty)
- `ublk_dirty_pages = 0` at the time of timeout — either the resync ran to completion but `__become_clean()` failed to update the route, or the metric is stale (showing the first resync's last value because `__run()` was never called in the second resync)
- The `write_superblock()` failure path in `__become_clean()` is NOT the cause — failures are logged as warnings and the route CAS proceeds regardless
- The TOCTOU race theory (a write dirtying the bitmap between `__run()` returning and `complete()` being called at line 120) is **not valid**: any write with `_outstanding_writes == 0` blocks in `enqueue_write()` → `__pause()` waiting for `SLEEPING` state, which never comes after `__run()` exits — `__pause()` only unblocks at `IDLE` (line 124), after `complete()` has already been called

## Root Cause

**Unknown — under investigation.**

## Files of Interest

- `src/raid/raid1/raid1_resync_task.cpp` — `_start()`, `__run()`
- `src/raid/raid1/raid1.cpp` — `swap_device()`, `toggle_resync()`, `__become_clean()`
- `src/raid/raid1/bitmap.cpp` — `dirty_pages()`, `dirty_region()`

szmyd:
# `stop()` IDLE→STOPPING race hangs resync after partition move

_type_: 🐛 (bug)
_severity_: 🔴 (critical — resync never completes, array stuck degraded indefinitely)
_discovery_: 🧪 (CI / functional test: `partition-move-2-0-nvme cold_with_io` timeout at 12000s)

1. **`stop()` CAS's `IDLE→STOPPING` when the thread has already finished, leaving state stuck.**
   The `IDLE+joinable` branch in `stop()` covers two distinct sub-cases that cannot be
   distinguished by state alone:

   | Case | State | Thread running? | Needs `CAS(IDLE→STOPPING)`? |
   |---|---|---|---|
   | Thread in unavail-wait | `IDLE` | Yes | **Yes** — to signal exit |
   | Thread just finished | `IDLE` | No (returning) | No — but gets CAS'd anyway |

   In the unavail-wait case the resync thread is alive but stuck in `_start()`'s top loop
   because `dirty_mirror->unavail` is set; the `CAS(IDLE→ACTIVE)` never fires so state stays
   `IDLE`. The `RETRY_WITH_SLEEP` in `stop()` is *correct and necessary* here — it eventually
   drives `CAS(IDLE→STOPPING)` which the thread sees at `__load_state()` and uses to exit.

   In the "just finished" case the thread already ran `xchng_status(ACTIVE→IDLE)` and returned.
   `stop()` still CAS's `IDLE→STOPPING` (via the same `RETRY_WITH_SLEEP`), joins the already-done
   thread immediately, then returns — leaving state permanently at `STOPPING` with no thread to
   clear it. The subsequent `toggle_resync(true)` → `launch()` call spins forever in
   `__transition_to` waiting for `STOPPING→IDLE` that never comes. `swap_device()` never returns;
   the second resync is never started.

   Observed as `active+dirty+syncing` indefinitely with `ublk_dirty_pages=0` (stale from the
   first resync — `__run()` was never entered for the second resync).

   **Fix:** keep `RETRY_WITH_SLEEP` (needed for the unavail-wait case). Add a post-join
   `CAS(STOPPING→IDLE)` at the end of `stop()` after the `join()`. This CAS is a no-op when
   the thread cleared `STOPPING` itself (normal stop path), and rescues the orphaned `STOPPING`
   state when the thread was already gone before `stop()` ran.

## Notes

Closes #205.

The two-commit structure is intentional: the first commit adds the failing regression test
(`Raid1Concurrency.StopAndRelaunchAfterComplete`) and the second applies the fix. CI failure on
commit 1 proves the test catches the bug; CI green on commit 2 proves the fix resolves it.

### _Additional cleanup_

- `StopAndRelaunchAfterComplete` regression test added to `resync_relaunch_after_complete.cpp`.
  Runs the second `launch()` in a thread with a 2s deadline; detaches and `FAIL()`s cleanly if
  hung rather than blocking CI indefinitely. `task` is heap-allocated via `shared_ptr` so the
  detached thread never accesses freed memory in the failing case.
- Version bumped to `0.22.2`, CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.22.2
+- raid1: Fix `stop()` IDLE→STOPPING race — when the resync thread finishes naturally and
+  `stop()` is called before `join()`, the `IDLE+joinable` handler now returns `SUCCESS` instead
+  of `RETRY_WITH_SLEEP`, preventing an accidental `CAS(IDLE→STOPPING)` that left no thread to
+  clear the state; subsequent `launch()` call in `swap_device()` would spin forever (SDSTOR-21927)
+
 ## 0.22.1
 - raid1: Fix dequeue/resume race — `_resync_state` and `_outstanding_writes` are now packed into a single `sisl::atomic_status_counter` so the counter decrement and PAUSE→ACTIVE transition are one indivisible CAS; `__resume()` is removed (SDSTOR-21927)
 - raid1: Fix enqueue/pause race — `enqueue_write()` now always calls `__pause()` on every enqueue, not only the first; previously a concurrent second enqueuer could skip `__pause()` while the first was still establishing it, allowing resync to overwrite an in-flight write with stale data
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - raid1: Fix `stop()` IDLE→STOPPING race — when the resync thread finishes naturally and
   `stop()` is called before `join()`, the `IDLE+joinable` handler now returns `SUCCESS` instead
   of `RETRY_WITH_SLEEP`, preventing an accidental `CAS(IDLE→STOPPING)` that left no thread to
-  clear the state; subsequent `launch()` call in `swap_device()` would spin forever (SDSTOR-21927)
+  clear the state; subsequent `launch()` call in `swap_device()` would spin forever.
 
 ## 0.22.1
-- raid1: Fix dequeue/resume race — `_resync_state` and `_outstanding_writes` are now packed into a single `sisl::atomic_status_counter` so the counter decrement and PAUSE→ACTIVE transition are one indivisible CAS; `__resume()` is removed (SDSTOR-21927)
+- raid1: Fix dequeue/resume race — `_resync_state` and `_outstanding_writes` are now packed into a single `sisl::atomic_status_counter` so the counter decrement and PAUSE→ACTIVE transition are one indivisible CAS; `__resume()` is removed.
 - raid1: Fix enqueue/pause race — `enqueue_write()` now always calls `__pause()` on every enqueue, not only the first; previously a concurrent second enqueuer could skip `__pause()` while the first was still establishing it, allowing resync to overwrite an in-flight write with stale data
 - raid1: Replace GCC `__builtin_popcount`/`__builtin_clz`/`__builtin_ctz` with C++23 `std::popcount`/`std::countl_zero`/`std::countr_zero`
 - build: `libatomic` is now declared as a Conan system lib on Linux — propagated automatically to consumers, no downstream changes required
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=2.0"
 
 class UBlkPPConan(ConanFile):
     name = "ublkpp"
-    version = "0.22.1"
+    version = "0.22.2"
 
     homepage = "https://github.com/szmyd/ublkpp"
     description = "A UBlk library for CPP application"
--- a/src/raid/raid1/raid1_resync_task.cpp
+++ b/src/raid/raid1/raid1_resync_task.cpp
@@ -284,6 +284,11 @@ void Raid1ResyncTask::stop() noexcept {
         std::unreachable();
     });
     if (_resync_task.joinable()) _resync_task.join();
+    // If the thread finished naturally (ACTIVE→IDLE) before stop() CAS'd IDLE→STOPPING,
+    // it returned without ever seeing STOPPING and never cleared it. _launch_lock is held
+    // so no concurrent caller can observe this window; reset to IDLE so launch() isn't stuck.
+    auto stopping = resync_state::STOPPING;
+    __cas_state(stopping, resync_state::IDLE);
 }
 
 resync_state Raid1ResyncTask::__yield(std::chrono::microseconds const yield_for,
--- a/src/raid/raid1/tests/concurrency/resync_relaunch_after_complete.cpp
+++ b/src/raid/raid1/tests/concurrency/resync_relaunch_after_complete.cpp
@@ -70,3 +70,89 @@ TEST(Raid1Concurrency, ResyncRelaunchAfterComplete) {
 
     task.stop();
 }
+
+// Regression test for SDSTOR-21927 / GH #205.
+//
+// Root cause: stop() CAS's IDLE→STOPPING when the resync thread has already finished
+// (state=IDLE, thread still joinable). State gets stuck at STOPPING with no running thread
+// to advance it back. The next launch() call spins forever in __transition_to.
+//
+// This sequence occurs in swap_device():
+//   toggle_resync(false)  [stop hits IDLE+joinable → CAS IDLE→STOPPING]
+//   __swap_device()       [sets route≠EITHER, dirties bitmap]
+//   toggle_resync(true)   [launch() loops forever on STOPPING]
+//
+// Without the fix: launch() spins forever and the test fails after 2s.
+// With the fix: stop() returns SUCCESS from IDLE+joinable (no CAS), state stays IDLE,
+// and the second launch() starts the new resync normally.
+//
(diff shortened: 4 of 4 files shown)


szmyd:
Without fix (4fd1d40):
```[ RUN      ] Raid1Concurrency.StopAndRelaunchAfterComplete
Warning:  19:46:17.984] [/home/runner/work/ublkpp/ublkpp/build/Release/src/raid/raid1/tests/test_raid1] [warning] [4037] [raid1_superblock.cpp:89:load_superblock] Missing superblock from: [DiskA, size=1024Mi, lbs=0x1000]
Warning:  19:46:17.984] [/home/runner/work/ublkpp/ublkpp/build/Release/src/raid/raid1/tests/test_raid1] [warning] [4037] [raid1_superblock.cpp:89:load_superblock] Missing superblock from: [DiskB, size=1024Mi, lbs=0x1000]
/home/runner/work/ublkpp/ublkpp/src/raid/raid1/tests/concurrency/resync_relaunch_after_complete.cpp:147: Failure
Failed
launch() hung after stop()+relaunch — SDSTOR-21927 IDLE→STOPPING race in stop()

[  FAILED  ] Raid1Concurrency.StopAndRelaunchAfterComplete (2005 ms)
```

shosseinimotlagh (approved):
LGTM

shosseinimotlagh:
Code Quality Assessment

  ✅ Strengths

  1. Excellent PR structure: Three-commit approach (test → fix → cleanup) makes the fix verifiable and reviewable
  2. Outstanding documentation: The PR description provides a detailed root cause analysis with a clear table explaining the two sub-cases
  3. Comprehensive regression test: The new test specifically reproduces the race condition with proper timeout handling
  4. Minimal, surgical fix: Only 5 lines of production code changed
  5. Clean test design: Uses shared_ptr to prevent UAF when the test times out and detaches the stuck thread

  🎯 The Core Fix

  Location: src/raid/raid1/raid1_resync_task.cpp:287-291

  +    // If the thread finished naturally (ACTIVE→IDLE) before stop() CAS'd IDLE→STOPPING,
  +    // it returned without ever seeing STOPPING and never cleared it. _launch_lock is held
  +    // so no concurrent caller can observe this window; reset to IDLE so launch() isn't stuck.
  +    auto stopping = resync_state::STOPPING;
  +    __cas_state(stopping, resync_state::IDLE);

  Analysis: The fix is elegant and correct. It adds a cleanup CAS after the join that:
  - Rescues orphaned STOPPING state when the thread finished before stop() ran
  - Is a no-op in the normal case (thread cleared STOPPING itself)
  - Is safe because _launch_lock is held, preventing concurrent observers

  ---
  Specific Suggestions

  1. Minor: Consider using compare_exchange_strong return value

  The current code doesn't check if the CAS succeeded, which is fine for this cleanup case. However, for debugging purposes, you might want to log when this rescue path triggers:

  auto stopping = resync_state::STOPPING;
  if (__cas_state(stopping, resync_state::IDLE)) {
      // Log or increment metric: "Rescued orphaned STOPPING state"
  }

  Rationale: This would help track how often this edge case occurs in production.

  2. Test refinement: Magic number 5ms sleep

  In the test (line 124):
  std::this_thread::sleep_for(5ms);

  Suggestion: Add a comment explaining why 5ms was chosen, or consider making it a named constant:

  // Give enough time for resync thread to execute ACTIVE→IDLE CAS and return,
  // but before stop() joins, to hit the IDLE+joinable race window.
  constexpr auto kRaceWindowDelay = 5ms;
  std::this_thread::sleep_for(kRaceWindowDelay);

  3. Test coverage: Verify state is IDLE after stop()

  Consider adding an assertion after task->stop() to verify the fix:

  task->stop();
  // Verify stop() cleaned up orphaned STOPPING state
  EXPECT_EQ(task->state(), resync_state::IDLE); // If task exposes state()

  Note: Only if the task exposes state inspection for testing.

  ---
  Potential Issues & Risks

  ⚠️ Low Risk: ABA Problem

  The cleanup CAS could theoretically fail if:
  1. Thread is at STOPPING when stop() joins
  2. Concurrent operation transitions STOPPING→IDLE→ACTIVE between the join and cleanup CAS

  Analysis: This is prevented by _launch_lock being held during stop(), as noted in the comment. ✅ Safe.

  ⚠️ Question: Why not fix the root cause?

  The root cause is that the IDLE+joinable case conflates two scenarios. An alternative fix would be to check if the thread is actually running before CAS'ing IDLE→STOPPING:

  // Pseudo-code alternative approach
  if (state == IDLE && joinable) {
      if (thread_still_running()) {
          return RETRY_WITH_SLEEP;  // Case 1: unavail-wait
      } else {
          return SUCCESS;  // Case 2: already finished
      }
  }

  Counter-argument: There's no portable way to check if a std::thread is running without joining it. The current fix is cleaner and more robust.

  ---
  Security & Correctness

  - ✅ No security implications — internal state machine fix
  - ✅ Thread safety: Protected by _launch_lock
  - ✅ Memory safety: Test uses shared_ptr to avoid UAF on detached thread
  - ✅ Atomic operations: Proper memory ordering (memory_order_acquire/release in tests)

  ---
  Test Coverage

  ✅ Excellent Test Design

  1. Reproduces the exact scenario: First resync completes → stop() → second launch()
  2. Detects the hang: 2-second timeout with detach to avoid blocking CI
  3. Avoids UB: shared_ptr<Raid1ResyncTask> ensures detached thread doesn't access freed memory
  4. Verifies both resyncs complete: Checks first_complete and second_complete atomics

  Suggested Additional Test Cases

  Consider adding tests for:
  1. Concurrent stop() calls (if not already covered)
  2. stop() during unavail-wait (case 1 from the table) — verify RETRY_WITH_SLEEP still works
  3. Multiple swap_device() cycles — stress test the pattern

  ---
  Performance Implications

  - ✅ Minimal overhead: One extra CAS operation per stop() call
  - ✅ Hot path unaffected: No changes to launch(), _start(), or I/O path

  ---
  Documentation & Changelog

  ✅ Strengths

  - CHANGELOG entry is clear and mentions the version bump
  - Commit messages are detailed and reference both internal ticket (SDSTOR-21927) and GitHub issue (#205)
  - Inline comments in the fix are excellent

  📝 Minor Suggestion

  The CHANGELOG description could mention the user-visible symptom:

  - raid1: Fix `stop()` IDLE→STOPPING race — when the resync thread finishes naturally and
    `stop()` is called before `join()`, the `IDLE+joinable` handler now returns `SUCCESS` instead
    of `RETRY_WITH_SLEEP`, preventing an accidental `CAS(IDLE→STOPPING)` that left no thread to
  - clear the state; subsequent `launch()` call in `swap_device()` would spin forever.
  + clear the state; subsequent `launch()` call in `swap_device()` would spin forever, leaving
  + the RAID array stuck in degraded mode indefinitely with resync never completing.

  ---
  Final Verdict

  ✅ APPROVE with minor suggestions

  This is a high-quality fix for a critical concurrency bug:
  - The root cause analysis is thorough and well-documented
  - The fix is minimal, correct, and well-commented
  - The regression test effectively catches the bug
  - The three-commit structure provides excellent reviewability

  The suggestions above are optional improvements — the PR is ready to merge as-is.

  Recommendation

  - ✅ Merge after CI passes
  - 📋 Consider backporting to stable branches (if applicable)
  - 📊 Monitor production metrics to see if the rescue CAS ever triggers

  Great work on catching and fixing this subtle race condition! The 5ms sleep in the test cleverly hits the exact race window. 🎯

PR #212, opened 2026-04-29
Feature/210 p1 custom cqe

### LINKED ISSUE #218: Phase 1 — Per-sub_cmd CqeState pool
Part of #210.

## What was done

Introduced a per-sub_cmd `CqeState` pool inside `async_io`, replacing the shared `tgt_io_cqe` / `async_completion` side-channel. `CqeAwaiter` added as the resume mechanism. `user_data` encoding and `tgt_io_done` routing unchanged.

Public header `include/ublkpp/lib/cqe_state.hpp` added for out-of-repo disk implementations.

## PR

#212

## Status

✅ Done

szmyd:
### Problem

The previous `tgt_io_done` / `handle_event` dispatch relied on a shared nullable-pointer side-channel (`tgt_io_cqe` / `async_completion`) that had to be written immediately before each coroutine resume. This meant only one sub_cmd could be in-flight per IO tag at a time — the pointer was a single slot, not a queue — and callers had to carefully time writes to avoid racing the resume.

### What this PR does

Replaces the side-channel with a per-sub_cmd `CqeState` pool (`std::deque<CqeState>`) inside each `async_io` slot. `build_cqe_state_data` registers a `CqeState` in the pool via `ensure(sub_cmd)` before submitting each SQE. When the CQE arrives, `tgt_io_done` looks up the right state by sub_cmd, writes the result, and calls `push_completion` to resume the coroutine.

`CqeAwaiter` replaces the raw coroutine handle store: it suspends until `completions` is non-empty, then pops and returns the next `CqeState*`. This correctly handles back-to-back CQEs without losing completions.

The public header `include/ublkpp/lib/cqe_state.hpp` is introduced so out-of-repo disk implementations can call `build_cqe_state_data` without depending on target internals.

### What is unchanged

`user_data` encoding (`tag | op | sub_cmd | bit63`) and all `tgt_io_done` / `handle_event` routing through ublksrv are preserved. Phase 2 replaces those.

### Test plan
- [x] `test_cqe_state`: pool creation, idempotency, FIFO ordering, waiter resume
- [x] Functional FSDisk read/write/flush/discard via MockUblksrv

### DIFF
--- a/include/ublkpp/lib/cqe_state.hpp
+++ b/include/ublkpp/lib/cqe_state.hpp
@@ -1,0 +1,85 @@
+#pragma once
+
+#include <coroutine>
+#include <cstdint>
+#include <deque>
+#include <utility>
+
+#include <sisl/logging/logging.h>
+#include <ublksrv.h>
+
+#include <ublkpp/lib/sub_cmd.hpp>
+
+namespace ublkpp {
+
+// Encodes tag, op, and sub_cmd into the io_uring SQE user_data for target I/O.
+// The high bit marks the entry as a target SQE so libublksrv routes the CQE to tgt_io_done.
+inline uint64_t build_tgt_sqe_data(uint64_t tag, uint64_t op, uint64_t sub_cmd) {
+    DEBUG_ASSERT_LE(tag, UINT16_MAX, "Tag too big: [{:#0x}]", tag)
+    DEBUG_ASSERT_LE(op, UINT8_MAX, "Op too big: [{:#0x}]", op)
+    DEBUG_ASSERT_LE(sub_cmd, UINT16_MAX, "SubCmd too big: [{:#0x}]", sub_cmd)
+    return tag | (op << sqe_tag_width) | (sub_cmd << (sqe_tag_width + sqe_op_width)) |
+        (static_cast< uint64_t >(0b1) << (sqe_tag_width + sqe_op_width + sqe_tgt_data_width + sqe_reserved_width));
+}
+
+struct CqeState;
+
+// Per-IO state tracking one inflight request, owned by the ublksrv-allocated io_data slot.
+//
+// Lifetime: placement-new'd in init_queue for each tag slot; explicitly ~async_io() in deinit_queue.
+// pool and completions are cleared at the start of each new I/O in __handle_io_async.
+//
+// Dispatch protocol:
+//   1. queue_tgt_io calls build_cqe_state_data → ensure(sub_cmd) per SQE; pool grows.
+//   2. tgt_io_done / handle_event: ensure(sub_cmd) finds the state, sets result, push_completion.
+//   3. push_completion resumes the CqeAwaiter-suspended coroutine if one is waiting.
+//   4. CqeAwaiter::await_resume pops one entry from completions and returns it.
+//   5. process_result inspects the state and may add to pending (retry / internal).
+struct async_io {
+    std::coroutine_handle<> waiter{};
+    std::deque< CqeState > pool{};         // stable addresses: push_back never invalidates pointers
+    std::deque< CqeState* > completions{}; // ordered queue of states ready for process_result
+    uint32_t pending{0};
+    int ret_val{0};
+
+    // Returns the CqeState for sub_cmd, creating it on first call. O(N) scan; N is small in practice.
+    CqeState* ensure(sub_cmd_t sub_cmd);
+
+    // Enqueues a completed state and resumes the coroutine if it is suspended on CqeAwaiter.
+    void push_completion(CqeState* s) {
+        completions.push_back(s);
+        if (auto h = std::exchange(waiter, {})) h.resume();
+    }
+};
+
+// Tracks a single inflight sub_cmd registered by build_cqe_state_data. Stored in async_io::pool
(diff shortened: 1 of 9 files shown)


Ilyin-Ilya (commented):
## Review findings

### Bug: `ensure()` called outside try-block — `std::bad_alloc` escapes through C callback → `std::terminate`

In both `tgt_io_done` and `handle_event`, `io->ensure(cmd)` is called **before** the `try` block. `ensure` calls `pool.push_back`, which can throw `std::bad_alloc`. Because these are C callbacks registered with ublksrv, a thrown exception has no C++ frame to catch it and the process calls `std::terminate`.

```cpp
// tgt_io_done — ensure is outside the try:
auto* state = io->ensure(cmd);   // <-- bad_alloc escapes here
state->result = cqe->res;
try {
    io->push_completion(state);
} catch (std::exception const& e) { ... }
```

Same pattern in `handle_event`.

**Fix:** move `ensure` inside the try block, or pre-allocate all `CqeState` entries in `build_cqe_state_data` only (which already does this) and make `ensure` in `tgt_io_done`/`handle_event` a non-allocating lookup that asserts the entry exists. The latter is safer because a missing entry at completion time indicates a logic error, not a recoverable OOM.

```cpp
static void tgt_io_done(ublksrv_queue const* q, ublk_io_data const* data, io_uring_cqe const* cqe) {
    RELEASE_ASSERT_EQ(data->tag, static_cast<int32_t>(user_data_to_tag(cqe->user_data)), "Tag mismatch!")
    auto* io = reinterpret_cast<async_io*>(data->private_data);
    auto cmd = static_cast<sub_cmd_t>(user_data_to_tgt_data(cqe->user_data));
    try {
        auto* state = io->ensure(cmd);   // moved inside
        state->result = cqe->res;
        io->push_completion(state);
        TLOGE("I/O threw exception: [{}]", e.what())
        ublksrv_complete_io(q, data->tag, -EIO);
    }
}
```

---

### Design: `async_io` full definition in public header unnecessarily exposes coroutine internals

`include/ublkpp/lib/cqe_state.hpp` is a public header intended to let out-of-repo drivers call `build_cqe_state_data`. The PR description says the full `async_io` definition is needed for that — but `build_cqe_state_data` only calls `async_io::ensure` through a pointer obtained from `data->private_data`. A forward declaration of `async_io` in the public header is sufficient; the full definition (including `waiter`, `completions`, `push_completion`, etc.) can remain in a private/internal header. This keeps the coroutine scheduling details out of the ABI-visible surface area.

---

### Design: Reset logic duplicated and diverged between `__handle_io_async` and `mock_ublksrv::submit_io`

`__handle_io_async` resets all five `async_io` fields at IO start:
```cpp
io->pool.clear(); io->completions.clear(); io->waiter = {}; io->ret_val = -EIO; io->pending = 0;
```

`mock_ublksrv::submit_io` resets only three:
```cpp
_io_states[tag].pool.clear(); _io_states[tag].completions.clear(); _io_states[tag].waiter = {};
```

`ret_val` and `pending` are not reset in the mock. The mock doesn't drive the coroutine so this doesn't currently cause visible failures, but it creates a maintenance trap: if the mock is ever extended to use `__handle_io_async`, stale `ret_val`/`pending` will corrupt the first IO on a reused tag slot. Extract an `async_io::reset()` method and call it from both sites.

---

### Naming conventions

Per the project's conventions (see CLAUDE.md), structs use `snake_case` and member variables carry a leading `_` prefix:

- `CqeState` → `cqe_state`; `CqeAwaiter` → `cqe_awaiter`
- Members `waiter`, `pool`, `completions`, `pending`, `ret_val`, `owner`, `result`, `sub_cmd` should be `_waiter`, `_pool`, etc.

Existing types `async_io` and `co_io_job` follow the convention correctly.

---

### Minor: `build_cqe_state_data` parameter type mismatch

`build_cqe_state_data` declares `uint64_t sub_cmd` but immediately casts it to `sub_cmd_t` for `ensure`. The parameter should be typed `sub_cmd_t` directly (matching `ensure`'s signature) so the compiler enforces the constraint at the call site rather than silently truncating.

szmyd:
Thanks for the thorough review. These are all valid observations -- responding to each with where they land in the stacked phases:

**Bug: `ensure()` outside try-block**
Moot after P6 (#224): `tgt_io_done` and `handle_event` are removed entirely. `__handle_io_async` becomes the sole coroutine entry point and does its own error handling inline.

**Design: `async_io` full definition in public header**
Resolved across P7-P8: `build_cqe_state_data` is deleted in P7 (#228) along with the entire `sub_cmd_t` system. By P8 (#230), `ensure(sub_cmd)` becomes the allocation-free `next_state()`; the `async_io` struct shrinks to just what the target layer needs and the public header concern dissolves.

**Design: reset logic duplicated between `__handle_io_async` and mock**
Resolved in P6 (#224): `tgt_io_done`/`handle_event` are gone; `__handle_io_async` is the single reset + drive site. The mock divergence disappears because the mock is updated to match in P5 (#217).

**Naming conventions (`CqeState`, member `_` prefix)**
Valid ongoing debt -- `CqeState` and its members carry this naming through all 10 phases. Noted for a follow-on cleanup pass.

**`build_cqe_state_data` parameter type mismatch**
Removed in P7 (#228) along with `sub_cmd_t` and `build_cqe_state_data` themselves.

Issue #210, opened 2026-04-29
Replace hand-rolled continuation scheme with C++20 coroutines for readable, locally-verifiable I/O logic


szmyd:
## Implementation Phases

| Phase | Status | PR | Issue | Description |
|---|---|---|---|---|
| 1 - Per-sub_cmd CqeState pool | ✅ Done | #212 | #218 | `CqeState` pool in `async_io`, `CqeAwaiter`, eliminate shared `tgt_io_cqe`/`async_completion` side-channel. Public header `include/ublkpp/lib/cqe_state.hpp`. |
| 2 - Own the CQE processing loop | ✅ Done | #213 | #219 | ublksrv patch adds `ublksrv_handle_cmd_cqe` + `ublksrv_queue_update_idle`. Custom `run_queue_loop` replaces `ublksrv_process_io`. |
| 3 - Raw CqeState pointer in user_data | ✅ Done | #215 | #220 | `build_cqe_state_data` returns `CqeState* | bit63`. `run_queue_loop` decodes with a single pointer cast. `build_tgt_sqe_data` and packed encoding removed. |
| 4 - `disk_task<int>` async API (FSDisk + RAID0) | ✅ Done | #216 | #221 | `disk_task<T>` coroutine + `uses_async_api()` + `handle_io_async()`. `CqeAwaitable` per-state resume. `MockUblksrv` extended with `inject_cqe`; new `test_raid0_async` binary (10 tests). |
| 5 - RAID1 async API migration | ✅ Done | #217 | #222 | `handle_iov_async` virtual + `StartedTaskAwaitable<T>`. RAID0 fan-out rewritten. RAID1 async path: flush/read/write/discard. eventfd injection eliminated. RAID10 migrates automatically. New `test_raid1_async` binary (11 tests). |
| 6 - exec::async_scope + graceful shutdown | ✅ Done | #224 | #223 | Per-queue `exec::async_scope`. Graceful shutdown via drain loop. Removed `async_iov`, `on_io_complete`. |
[...]

PR #213, opened 2026-04-29
Feature/210 p2 own the cqe

### LINKED ISSUE #219: Phase 2 — Own the CQE processing loop
Part of #210.

## What was done

Patched ublksrv to expose two new public functions:

```c
void ublksrv_handle_cmd_cqe(const struct ublksrv_queue*, const struct io_uring_cqe*);
void ublksrv_queue_update_idle(const struct ublksrv_queue*, int ret, int reapped);
```

Replaced `ublksrv_process_io` with our own `run_queue_loop` over `ring_ptr`. Target CQEs dispatched inline (sub_cmd to CqeState routing preserved). `tgt_io_done` and `handle_event` removed from `tgt_type`.

Our loop:

```cpp
while (!ublksrv_queue_is_done(q)) {
    io_uring_cqe* cqe;
    auto ret = io_uring_submit_and_wait_timeout(ring_ptr, &cqe, 1, tsp, nullptr);
    unsigned head, count = 0;
    io_uring_for_each_cqe(ring_ptr, head, cqe) {
        if (cqe->user_data >> 63) {
            if (cqe->user_data >> 62 & 1) handle_event(q);
            else tgt_io_done_inline(q, cqe);
        } else {
            ublksrv_handle_cmd_cqe(q, cqe);
        }
        ++count;
    }
    io_uring_cq_advance(ring_ptr, count);
    ublksrv_queue_update_idle(q, ret, count);
}
```

## PR

#213

## Status

✅ Done

szmyd:
### Problem

`ublksrv_process_io` routes target CQEs through `tgt_io_done` and `handle_event` function pointers registered on `tgt_type`. This indirection exists solely to let the target intercept completions — but it means the dispatch path goes through ublksrv even though ublksrv adds no value there. More importantly, it prevents Phase 3: storing a raw `CqeState*` in `user_data` requires decoding it ourselves, which is impossible if ublksrv owns the loop.

### What this PR does

Patches ublksrv (version `nbi.1.5.0.1`) to expose two new public functions:

```c
void ublksrv_handle_cmd_cqe(const struct ublksrv_queue*, const struct io_uring_cqe*);
void ublksrv_queue_update_idle(const struct ublksrv_queue*, int ret, int reapped);
```

With those available, `run_queue_loop` replaces `ublksrv_process_io` entirely. It iterates the raw `io_uring` ring directly: target CQEs (bit 63 set) are dispatched inline, eventfd CQEs (bit 62 set) collect async results via `collect_async`, and ublk command CQEs (FETCH/COMMIT) delegate to `ublksrv_handle_cmd_cqe`. `tgt_io_done` and `handle_event` are set to `nullptr` on `tgt_type` — they are no longer called.

The sub_cmd→CqeState routing inside the target CQE branch is identical to Phase 1; only the outer dispatch mechanism changes.

MockUblksrv gains a proper `async_io` backing store per tag slot so `build_cqe_state_data` can dereference `private_data` safely in tests without the real ublksrv init path.

### What is unchanged

`user_data` encoding (`tag | op | sub_cmd | bit63`) and the per-sub_cmd pool lookup are preserved. Phase 3 replaces those.

### Test plan
- [x] All 13 existing tests pass (Debug + ASAN)
- [x] Functional FSDisk tests exercise `run_queue_loop` dispatch via MockUblksrv

### DIFF
--- a/3rd_party/ublksrv/patches/run_loop_1_5_0.patch
+++ b/3rd_party/ublksrv/patches/run_loop_1_5_0.patch
@@ -1,0 +1,134 @@
+diff --git a/include/ublksrv.h b/include/ublksrv.h
+index edaa4e9..7e58f2a 100644
+--- a/include/ublksrv.h
++++ b/include/ublksrv.h
+@@ -1155,6 +1155,21 @@ extern int ublksrv_complete_io(const struct ublksrv_queue *q, unsigned tag, int
+  */
+ extern int ublksrv_queue_reap_events(const struct ublksrv_queue *tq);
+ 
++/**
++ * ublksrv_handle_cmd_cqe - handle one ublk command CQE (non-target IO).
++ * Caller must verify is_target_io(cqe->user_data) == 0 before calling.
++ */
++extern void ublksrv_handle_cmd_cqe(const struct ublksrv_queue *tq,
++		const struct io_uring_cqe *cqe);
++
++/**
++ * ublksrv_queue_update_idle - update idle/stopping state after draining CQEs.
++ * @ret:    return value of io_uring_submit_and_wait_timeout
++ * @reapped: number of CQEs consumed in this batch
++ */
++extern void ublksrv_queue_update_idle(const struct ublksrv_queue *tq,
++		int ret, int reapped);
++
+ /** @} */ // end of ublksrv_queue group
+ 
+ #ifdef __cplusplus
+diff --git a/lib/ublksrv.c b/lib/ublksrv.c
+index e004b22..6472ebc 100644
+--- a/lib/ublksrv.c
++++ b/lib/ublksrv.c
+@@ -467,12 +467,6 @@ static int ublksrv_setup_eventfd(struct _ublksrv_queue *q)
+ 		return -EINVAL;
+ 	}
+ 
+-	if (!q->dev->tgt.ops->handle_event) {
+-		ublk_err("ublk dev %d/%d not define ->handle_event",
+-			info->dev_id, q->q_id);
+-		return -EINVAL;
+-	}
+-
+ 	q->efd = eventfd(0, 0);
+ 	if (q->efd < 0)
+ 		return -errno;
+@@ -829,30 +823,14 @@ static inline void ublksrv_handle_tgt_cqe(struct _ublksrv_queue *q,
+ 	}
+ }
+ 
+-static void ublksrv_handle_cqe(struct io_uring *r,
+-		struct io_uring_cqe *cqe, void *data)
++static void ublksrv_handle_cmd_cqe_impl(struct _ublksrv_queue *q,
++		const struct io_uring_cqe *cqe)
+ {
+-	struct _ublksrv_queue *q = container_of(r, struct _ublksrv_queue, ring);
+ 	unsigned tag = user_data_to_tag(cqe->user_data);
+-	unsigned cmd_op = user_data_to_op(cqe->user_data);
+ 	int fetch = (cqe->res != UBLK_IO_RES_ABORT) &&
+ 		!(q->state & UBLKSRV_QUEUE_STOPPING);
+-	struct ublk_io *io;
+-
+-	ublk_dbg(UBLK_DBG_IO_CMD, "%s: res %d (qid %d tag %u cmd_op %u target %d/%x event %d) stopping %d\n",
+-			__func__, cqe->res, q->q_id, tag, cmd_op,
+-			is_target_io(cqe->user_data),
+-			user_data_to_tgt_data(cqe->user_data),
+-			is_eventfd_io(cqe->user_data),
+-			(q->state & UBLKSRV_QUEUE_STOPPING));
+-
+-	/* Don't retrieve io in case of target io */
+-	if (is_target_io(cqe->user_data)) {
+-		ublksrv_handle_tgt_cqe(q, cqe);
+-		return;
+-	}
++	struct ublk_io *io = &q->ios[tag];
+ 
+-	io = &q->ios[tag];
+ 	q->cmd_inflight--;
+ 
+ 	if (!fetch) {
+@@ -885,6 +863,29 @@ static void ublksrv_handle_cqe(struct io_uring *r,
+ 	}
+ }
+ 
++static void ublksrv_handle_cqe(struct io_uring *r,
++		struct io_uring_cqe *cqe, void *data)
++{
++	struct _ublksrv_queue *q = container_of(r, struct _ublksrv_queue, ring);
++	unsigned tag = user_data_to_tag(cqe->user_data);
++	unsigned cmd_op = user_data_to_op(cqe->user_data);
++
++	ublk_dbg(UBLK_DBG_IO_CMD, "%s: res %d (qid %d tag %u cmd_op %u target %d/%x event %d) stopping %d\n",
++			__func__, cqe->res, q->q_id, tag, cmd_op,
++			is_target_io(cqe->user_data),
++			user_data_to_tgt_data(cqe->user_data),
++			is_eventfd_io(cqe->user_data),
++			(q->state & UBLKSRV_QUEUE_STOPPING));
++
++	/* Don't retrieve io in case of target io */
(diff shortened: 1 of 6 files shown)


3rd_party/ublksrv/patches/run_loop_1_5_0.patch
```diff
@@ -0,0 +1,134 @@
...
++ * @ret:    return value of io_uring_submit_and_wait_timeout
++ * @reapped: number of CQEs consumed in this batch
++ */
++extern void ublksrv_queue_update_idle(const struct ublksrv_queue *tq,
++		int ret, int reapped);
++
+ /** @} */ // end of ublksrv_queue group
+ 
+ #ifdef __cplusplus
+diff --git a/lib/ublksrv.c b/lib/ublksrv.c
```
Ilyin-Ilya, line 27: btw we can patch and make stubs for ublksrv calls. 
Something like 
if #MOCK
{
mock_impl
}
else
{
real_impl
}
Is it possible?

src/target/ublkpp_tgt.cpp
```diff
@@ -76,6 +77,74 @@ static void set_queue_thread_affinity(ublksrv_ctrl_dev const*) {
...
+                        } catch (std::exception const& e) {
+                            TLOGE("Async event threw exception: [{}]", e.what())
+                            ublksrv_complete_io(q, r.io->tag, -EIO);
+                        }
+                    }
+                } else {
+                    // target io_uring CQE — route by sub_cmd to CqeState
+                    auto* data =
+                        reinterpret_cast< ublk_io_data* >(ublksrv_io_private_data(q, user_data_to_tag(cqe->user_data)));
+                    RELEASE_ASSERT_EQ(data->tag, static_cast< int32_t >(user_data_to_tag(cqe->user_data)),
```
Ilyin-Ilya, line 124: **Bug: type confusion — `ublksrv_io_private_data` returns `async_io\*`, not `ublk_io_data\*`**

`ublksrv_io_private_data(q, tag)` is defined in `lib/ublksrv.c` as:
```c
void *ublksrv_io_private_data(const struct ublksrv_queue *tq, int tag) {
    return tq_to_local(tq)->ios[tag].data.private_data;
}
```
It returns `ublk_io_data::private_data` — the `async_io\*` buffer stored inside the io — **not** a pointer to `ublk_io_data` itself.

Casting this to `ublk_io_data\*` and then reading `data->tag` and `data->private_data` reads from the wrong struct entirely (interpreting `async_io` bytes as `ublk_io_data` fields). The `RELEASE_ASSERT_EQ(data->tag, ...))` immediately after will fire with garbage or silently pass with incorrect values.

The correct API to get a `const ublk_io_data\*` is `ublksrv_queue_get_io_data(q, tag)`. And `ublksrv_io_private_data` already gives you the `async_io\*` directly:

```cpp
// Correct fix:
auto tag = static_cast< int >(user_data_to_tag(cqe->user_data));
auto* data = ublksrv_queue_get_io_data(q, tag);  // returns const ublk_io_data*
RELEASE_ASSERT_EQ(data->tag, static_cast< int32_t >(tag), "Tag mismatch!")
auto* io = reinterpret_cast< async_io* >(data->private_data);
// -- or equivalently --
auto* io = static_cast< async_io* >(ublksrv_io_private_data(q, tag));
```

Note: `init_tgt` at line 473/479 uses `ublksrv_io_private_data` correctly (placing/destroying `async_io` in that buffer). This target CQE branch is the only place where the wrong cast is applied.

src/target/ublkpp_tgt.cpp
```diff
@@ -76,6 +77,74 @@ static void set_queue_thread_affinity(ublksrv_ctrl_dev const*) {
...
+        io_uring_for_each_cqe(ring, head, cqe) {
+            if (cqe->user_data & k_target_bit) {
+                if (cqe->user_data & k_eventfd_bit) {
+                    // eventfd notification — collect async completions (RAID1/iSCSI)
+                    auto completed = std::list< async_result >();
+                    tgt->device->collect_async(q, completed);
+                    ublksrv_queue_handled_event(q);
+                    for (auto& r : completed) {
+                        auto* io = reinterpret_cast< async_io* >(r.io->private_data);
+                        auto* state = io->ensure(r.sub_cmd);
```
Ilyin-Ilya, line 111: **Bug: `io->ensure()` called outside the `try` block — OOM exception escapes `io_uring_for_each_cqe`**

This was raised in PR #212 and is still present here (in both the eventfd branch and the target-CQE branch below at line 128).

If `io->ensure(r.sub_cmd)` throws `std::bad_alloc` (OOM), the exception propagates out of the `io_uring_for_each_cqe` macro before `io_uring_cq_advance` is called. The same batch of CQEs will be re-processed on the next `io_uring_submit_and_wait_timeout`, producing duplicate completions, double-frees, or an infinite loop.

Suggested fix — wrap both `ensure` and `push_completion` in the same try:
```cpp
try {
    auto* state = io->ensure(r.sub_cmd);
    state->result = r.result;
    io->push_completion(state);
    TLOGE("Async event threw exception: [{}]", e.what())
    ublksrv_complete_io(q, r.io->tag, -EIO);
}
```
The same fix applies to the target-CQE branch.

Ilyin-Ilya (commented):
**Design observations**

**D — Duplicated private constants with no compile-time guard**

`k_io_idle_secs = 20`, `k_target_bit = 1ULL << 63`, and `k_eventfd_bit = 1ULL << 62` are copied from ublksrv private internals (`ublksrv.c` / `ublksrv_priv.h`) with a comment acknowledging the duplication. If upstream changes any of these (e.g. repurposes bit 62 or adjusts the idle timeout), the code will silently misbehave.

Since this PR already patches ublksrv, consider exposing these as public constants in `include/ublksrv.h` within the patch and `#include`-ing them, or at minimum adding a static_assert or `__attribute__((deprecated))` sentinel that triggers a build break on mismatch.

**E — No direct tests for `run_queue_loop` dispatch branches**

The three CQE dispatch paths (target io_uring CQE, eventfd CQE, ublk command CQE) and the idle timeout path are exercised only indirectly through integration tests. The idle path (`k_io_idle_secs` timeout firing while ring is empty) appears to have no coverage at all. Given that this loop replaces `ublksrv_process_io` entirely, unit tests that inject synthetic CQEs and mock `ublksrv_handle_cmd_cqe` / `collect_async` would significantly reduce risk, especially around the ordering invariants.

**F — `ublksrv_setup_eventfd` null-check removal is undocumented**

The patch removes the `handle_event != nullptr` guard from `ublksrv_setup_eventfd`:
```c
-  if (!q->dev->tgt.ops->handle_event) {
-      ublk_err("ublk dev %d/%d not define ->handle_event", ...);
-      return -EINVAL;
-  }
```
This is correct for the ublkpp use-case (we now handle the event inline), but the guard was a safety net for *any* caller of `ublksrv_setup_eventfd`. With it gone, other targets that set `UBLKSRV_F_NEED_EVENTFD` but forget to register `handle_event` will no longer get an error at setup time — they'll set up an eventfd and then silently drop notifications. The patch description should note this behavioral change for downstream ublksrv users, and ideally ublksrv's own documentation should be updated.

szmyd:
Thanks -- addressing each finding against the stacked phases:

**D: Duplicated private constants**
`k_eventfd_bit` is removed entirely in P5 (#217) when the eventfd path is eliminated. `k_target_bit` and `k_io_idle_secs` remain in P10 but are isolated to a single file (`ublkpp_tgt.cpp`) with an explanatory comment; they never enter a shared header.

**E: No direct tests for `run_queue_loop` dispatch branches**
Acknowledged as a known gap -- see the follow-on comment in issue #210. The mock currently bypasses `handle_io_async` and calls down to the disk implementation directly. Wiring up the full coroutine dispatch path in the mock is tracked as future work after P10 stabilises.

**F: `ublksrv_setup_eventfd` null-check removal undocumented**
P5 (#217) removes the eventfd path from ublkpp entirely, so the patched behaviour is never exercised after that phase. The ublksrv patch change is permanent but its blast radius shrinks to zero for ublkpp's use-case.

PR #215, opened 2026-04-30
Feature/210 p3 Stash CqeState* in SQE directly.

### LINKED ISSUE #220: Phase 3 — Raw CqeState pointer in user_data
Part of #210.

## What was done

With our own CQE loop, `user_data` for target SQEs is entirely ours. `build_cqe_state_data` now stores a raw `CqeState*` (with bit 63 set as the target marker):

```cpp
inline uint64_t build_cqe_state_data(ublk_io_data const* data, uint64_t sub_cmd) {
    auto* state = reinterpret_cast<async_io*>(data->private_data)->ensure(sub_cmd);
    return reinterpret_cast<uint64_t>(state) | (1ULL << 63);
}
```

The loop resolves it with a single pointer cast — no tag extraction, no sub_cmd lookup, no pool scan:

```cpp
auto* state = reinterpret_cast<CqeState*>(cqe->user_data & ~k_target_bit);
state->result = cqe->res;
state->owner->push_completion(state);
```

`build_tgt_sqe_data` and the packed encoding removed. `sub_cmd_t` retained in `CqeState::sub_cmd` for `process_result` flag tests and the eventfd path.

## PR

#215

## Status

✅ Done

szmyd:
### Problem

Phase 2 owns the CQE loop but target SQEs still carry the packed `tag | op | sub_cmd | bit63` encoding inherited from ublksrv. When a target CQE arrives, `run_queue_loop` must decode the tag, walk to `ublksrv_io_private_data`, cast to `async_io*`, extract the sub_cmd, then call `ensure()` to scan the pool O(N) and re-find the `CqeState`. All of that work just to get back to something `build_cqe_state_data` already had a pointer to when the SQE was submitted.

### What this PR does

`build_cqe_state_data` now calls `ensure(sub_cmd)` as before (registering the `CqeState` in the pool), but instead of returning the packed integer it returns the pointer directly:

```cpp
return reinterpret_cast<uint64_t>(state) | (1ULL << 63);
```

Bit 63 marks it as a target SQE. On ARM64/x86_64, canonical userspace addresses are ≤48 bits, so bit 63 is always clear in any valid pointer — the OR is safe and reversible.

`run_queue_loop` decodes with a single cast:

```cpp
auto* state = reinterpret_cast<CqeState*>(cqe->user_data & ~k_target_bit);
state->result = cqe->res;
state->owner->push_completion(state);
```

Five lines of decode + an O(N) pool scan become one. `build_tgt_sqe_data` is removed along with its test suite. `async_io` gains a `tag` field set at IO start so error paths and the mock can recover the tag from `state->owner->tag` without touching ublksrv data structures.

### What is unchanged

`sub_cmd_t` is retained — it is still stored in `CqeState::sub_cmd` and read by `process_result` flag tests (`is_replicate`, `is_internal`, etc.) and RAID routing. The eventfd path (`collect_async` → `async_result::sub_cmd` → `ensure()`) is also unchanged. Phase 4 eliminates `sub_cmd_t` entirely.

### Test plan
- [x] `BuildCqeStateData`: bit 63 set, lower bits decode to correct `CqeState*`
- [x] All 13 tests pass (Debug + ASAN)
- [x] FSDisk functional tests exercise the pointer-decode path in `run_queue_loop`


src/target/ublkpp_tgt.cpp
```diff
@@ -84,8 +84,8 @@ static constexpr uint64_t k_eventfd_bit = 1ULL << 62;
 static constexpr int k_io_idle_secs = 20;
 
 // Our own CQE processing loop, replacing ublksrv_process_io.
-// Target CQEs are dispatched inline; ublk command CQEs delegate to ublksrv.
-// sub_cmd encoding in user_data is preserved (Phase 2 — no pointer encoding yet).
+// Target CQEs carry a raw CqeState* in bits 62:0 with bit 63 set; decoded via pointer cast.
+// Eventfd CQEs use bit 62; ublk command CQEs delegate to ublksrv.
```
Ilyin-Ilya, line 88: **Bug (flagged in PRs #212 and #213, still not fixed): `ensure` in the eventfd branch is called outside the `try` block — uncaught `std::bad_alloc` crashes the queue thread**

At line 111 (inside the `if (cqe->user_data & k_eventfd_bit)` branch):

```cpp
auto* state = io->ensure(r.sub_cmd);   // ← outside try — bad_alloc propagates uncaught
state->result = r.result;
try {
    io->push_completion(state);
} catch (std::exception const& e) { ... }
```

`ensure` calls `pool.push_back(CqeState{...})` on a `std::deque`. `push_back` can throw `std::bad_alloc`. Because the call is *before* the `try`, the exception escapes the `io_uring_for_each_cqe` loop and terminates the queue thread — the same crash scenario flagged in the two prior reviews.

The target path (directly below) avoided this by encoding the `CqeState*` at SQE submission time, so no allocation is needed at CQE dispatch. The eventfd path still allocates at the worst possible moment.

**Fix:** at minimum, move `ensure` inside the `try`:
```cpp
for (auto& r : completed) {
    auto* io = reinterpret_cast< async_io* >(r.io->private_data);
    try {
        auto* state = io->ensure(r.sub_cmd);   // inside try
        state->result = r.result;
        io->push_completion(state);
        TLOGE("Async event threw exception: [{}]", e.what())
        ublksrv_complete_io(q, r.io->tag, -EIO);
    }
}
```
Ideally, Phase 4 should eliminate the `ensure` call at dispatch time entirely by carrying the `CqeState*` through the eventfd channel the same way the target path now does.

src/tests/mock_ublksrv/mock_ublksrv.cpp
```diff
@@ -96,8 +97,9 @@ io_result MockUblksrv::submit_io(int tag, uint8_t op, uint64_t start_sector, uin
 }
 
 void MockUblksrv::process_cqe(io_uring_cqe* cqe, std::vector< Completion >& out) {
-    int const tag = static_cast< int >(user_data_to_tag(cqe->user_data));
-    auto const sub_cmd = static_cast< sub_cmd_t >(user_data_to_tgt_data(cqe->user_data));
+    auto* state = reinterpret_cast< CqeState* >(cqe->user_data & ~(1ULL << 63));
```
Ilyin-Ilya, line 100: **Bug: `process_cqe` dereferences a `CqeState*` decoded from `cqe->user_data`, but `submit_io` calls `pool.clear()` (line 85) without first draining all CQEs from the prior IO — leaving dangling pointers in the ring**

After this PR, `cqe->user_data` holds a raw `CqeState*` into `async_io::pool`. If a caller issues a second `submit_io` on the same tag before all CQEs from the previous IO have been processed (e.g. early exit after a partial poll), `pool.clear()` destroys the `CqeState` objects that are still referenced by outstanding CQEs. When `process_cqe` later decodes and dereferences that pointer, it is use-after-free.

This was not a problem before this PR because `user_data` was a packed integer (tag + sub_cmd) with no pointer lifetime dependency. The switch to raw pointers introduced this hazard silently.

**Under the old encoding** `process_cqe` could tolerate a stale CQE because decoding produced an integer, not a pointer. **Under the new encoding** any stale CQE is an immediate memory-safety violation.

**Fix options:**
1. Add an assertion in `submit_io` that `sub_cmds_remaining == 0` before clearing the pool, making the invariant explicit:
   ```cpp
   DEBUG_ASSERT(ts.sub_cmds_remaining == 0,
       "submit_io called with {} outstanding CQEs on tag {}", ts.sub_cmds_remaining, tag);
   _io_states[tag].pool.clear();
   ```
2. Drain all pending CQEs for the tag before clearing (more defensive, but changes mock semantics).

Option 1 is the right call for a test mock — it catches misuse immediately rather than producing silent corruption.

include/ublkpp/lib/cqe_state.hpp
```diff
@@ -68,18 +59,13 @@ inline CqeState* async_io::ensure(sub_cmd_t sub_cmd) {
...
-inline uint64_t build_cqe_state_data(ublk_io_data const* data, uint64_t tag, uint64_t op, uint64_t sub_cmd) {
-    reinterpret_cast< async_io* >(data->private_data)->ensure(static_cast< sub_cmd_t >(sub_cmd));
-    return build_tgt_sqe_data(tag, op, sub_cmd);
+// Stores the CqeState* directly in the SQE user_data, OR'd with bit 63 to mark it as a target SQE.
+// On ARM64/x86_64 canonical userspace addresses use ≤48 bits, so bit 63 is always zero in any
+// valid pointer — the OR is safe and reversible with & ~(1ULL << 63).
+inline uint64_t build_cqe_state_data(ublk_io_data const* data, uint64_t sub_cmd) {
+    auto* io = reinterpret_cast< async_io* >(data->private_data);
+    auto* state = io->ensure(static_cast< sub_cmd_t >(sub_cmd));
+    return reinterpret_cast< uint64_t >(state) | (1ULL << 63);
```
Ilyin-Ilya, line 68: **Missing assertion: no check that the `CqeState*` fits in 63 bits before OR'ing with the marker**

The comment above claims canonical userspace addresses use ≤48 bits, so bit 63 is always clear. That is true on current x86_64 and ARM64 hardware. However there is no `DEBUG_ASSERT` (or `static_assert`) to catch the day this assumption breaks — whether from a future 5-level paging kernel (57-bit VA), an ASLR quirk in a test environment, or a bug in `ensure` returning a bogus pointer.

Additionally, the comment in `run_queue_loop` describes the encoding as "bits 62:0" which is actively misleading: `k_eventfd_bit = 1ULL << 62` uses bit 62 as a *separate* flag for eventfd CQEs, so bits 62:0 are not all available for the pointer. The comment should say "bits 61:0" or, more precisely, that canonical userspace pointers use at most 47 bits (user-space VA on current Linux), leaving bits 63:47 clear.

Neither of these issues causes a real bug on current hardware with current kernels, but the missing assertion means a future regression would manifest as silent memory corruption rather than an immediate caught failure.

**Suggested fix:**
```cpp
inline uint64_t build_cqe_state_data(ublk_io_data const* data, uint64_t sub_cmd) {
    auto* io = reinterpret_cast< async_io* >(data->private_data);
    auto* state = io->ensure(static_cast< sub_cmd_t >(sub_cmd));
    DEBUG_ASSERT(!(reinterpret_cast< uintptr_t >(state) & (1ULL << 63)),
        "CqeState* has bit 63 set — pointer encoding is unsafe")
    return reinterpret_cast< uint64_t >(state) | (1ULL << 63);
}
```
And fix the comment in `run_queue_loop` from "bits 62:0" to "bits 46:0 on current hardware (48-bit canonical VA)".

Ilyin-Ilya (commented):
**Design findings: comment accuracy, teardown ordering, naming convention**

---

**D) Comment "bits 62:0" in `run_queue_loop` is misleading**

The comment at the top of `run_queue_loop` says:
> Target CQEs carry a raw CqeState* in bits 62:0 with bit 63 set

But `k_eventfd_bit = 1ULL << 62` uses bit 62 as a *separate* flag for eventfd CQEs. A target CQE pointer therefore cannot safely occupy bit 62. The comment should say "bits 46:0 on current hardware (48-bit canonical VA)" — which is what the PR description actually says. The mismatch between the PR description and the in-code comment is a readability hazard for the next person modifying this dispatch loop.

**Suggested fix:** Change the comment to:
```
// Target CQEs carry a raw CqeState* in bits 46:0 (48-bit canonical VA) with bit 63 set; decoded via pointer cast.
```

---

**E) No comment or assertion establishing that all target CQEs are drained before `deinit_queue` destructs `async_io`**

`deinit_queue` calls `~async_io()` on every IO slot, which destructs `pool` (the slab backing all `CqeState` objects). If any io_uring target CQE is still in the ring at that point, its `user_data` holds a `CqeState*` that is now dangling — reading it is UB.

Under normal shutdown (`ublksrv_ctrl_stop_dev` → drain → `ublksrv_queue_is_done`) this is safe by construction, but that invariant is invisible from the code. A comment on `deinit_queue` (or a `DEBUG_ASSERT` that the ring is empty) would make the teardown contract explicit and catch future shutdown-path regressions early.

---

**F) `async_io::tag` violates the `_snake_case` member naming convention**

Per project conventions, member variables are named `_snake_case` (e.g. `_device`, `_ring`). The `tag` field added in this PR (and the associated `MockTagState::sub_cmds_remaining`, `result`, etc.) uses plain `snake_case` without a leading underscore. This is pre-existing debt from the struct's origin in prior PRs, but it is perpetuated here with the new `tag` field.

Not blocking, but worth noting for consistency. If a cleanup pass is planned, `async_io` is the place to start.

szmyd:
Thanks -- addressing each finding:

**D: Comment "bits 62:0" is misleading**
Correct for P3 where `k_eventfd_bit = 1ULL << 62` exists alongside `k_target_bit`. P5 (#217) removes `k_eventfd_bit` entirely, so by P10 (#234) bit 62 is genuinely available to the pointer and the comment becomes accurate. The misleading period is P3 only.

**E: No comment establishing teardown ordering for `deinit_queue`**
P10 (#234) replaces the drain loop with `co_await scope.on_empty()` and adds an explicit correctness argument in the comment above `run_queue_loop`: `ublksrv_queue_is_done` returns true only after all ublksrv I/Os are acknowledged, and we call `ublksrv_complete_io` at the end of `__handle_io_async` after the coroutine fully completes -- so the scope is empty when the loop exits and `deinit_queue` runs after that.

**F: `async_io::tag` naming convention**
Pre-existing debt carried through P10 unchanged. Noted for a future cleanup pass alongside the `CqeState` member naming. Addressed in P11 (#235)

PR #253, opened 2026-05-07
raid1: replace global PAUSE with lock-free per-region write tracker

### LINKED ISSUE #179: RAID1: Implement per-region tracking for resync-write coordination
## Summary

Implement **per-region tracking** to enable RAID1 resync to continue on non-conflicting regions while only pausing for actual overlapping I/O. Current global mutual exclusion is correct but inefficient - resync pauses for ALL writes even when they occur on completely different regions.

**Expected Performance Gain**: 10-50x faster resync under concurrent write load

## Problem

Current implementation uses global atomic counter (`_outstanding_writes`) for resync-write coordination:
- When ANY write is active, the ENTIRE resync task pauses
- On a 2 TB volume with random writes during resync, resync may be paused 90%+ of the time
- Write to region A pauses resync of region B (unnecessary)

### Design Options: Memory vs Latency Trade-off

Both options are **fully lock-free** and maintain correctness. The choice is between memory efficiency and conflict detection latency.

---

#### **Option 1: Lock-Free Bounded Array (RECOMMENDED)**

Track only active in-flight writes in a small fixed-size array, search linearly for conflicts.

```cpp
class RegionTracker {
    struct WriteRegion {
        std::atomic<uint64_t> addr;  // 0 = unused slot
        std::atomic<uint32_t> len;
    };
    
    std::vector<WriteRegion> _active_writes;  // size = qdepth (128 default)
    std::atomic<uint32_t> _active_count{0};
    
    void set_write_in_flight(uint64_t addr, uint32_t len) {
        // Find empty slot via CAS
        for (auto& slot : _active_writes) {
            uint64_t expected = 0;
            if (slot.addr.compare_exchange_strong(expected, addr, memory_order_release)) {
                slot.len.store(len, memory_order_release);
                return;
            }
        }
    }
    
    bool has_conflict(uint64_t addr, uint32_t len) {
        // Linear search through active writes
        for (const auto& slot : _active_writes) {
            uint64_t w_addr = slot.addr.load(memory_order_acquire);
            if (w_addr == 0) continue;
            
            uint32_t w_len = slot.len.load(memory_order_acquire);
            if (ranges_overlap(addr, len, w_addr, w_len)) return true;
        }
        return false;
    }
};
```

**Memory Footprint**:
- **1.5 KiB** (128 slots × 12 bytes)
- Bounded by queue depth, not volume size
- 170,000x less memory than Option 2

**Performance Characteristics**:

| Operation | Latency | Notes |
|-----------|---------|-------|
| `set_write_in_flight()` | ~50-200 ns | Average case: find empty slot in ~5 iterations |
| `has_conflict()` (resync) | ~100-500 ns | Linear search ~50 slots (typical), 128 max |
| `clear_write_in_flight()` | ~50-200 ns | Same as set (find matching slot) |

**Honest Latency Penalties**:
- ❌ **O(n) conflict detection**: Each resync chunk check scans ~50 active writes
  - 100 MB resync region = 3,200 chunks × 500 ns = **1.6 ms overhead**
  - Context: Reading 100 MB from NVMe takes ~200 ms, so overhead is **0.8%**
- ❌ **Worst case**: If queue depth is maxed (128 concurrent writes), search degrades to ~500 ns per check
- ❌ **Cache thrashing**: Under extreme load, iterating hot atomic variables may cause cache coherence traffic

**Benefits**:
- ✅ Tiny memory footprint (1.5 KiB vs 256 MB)
- ✅ Simple implementation (no complex indexing logic)
- ✅ No slot exhaustion issues (bounded by queue depth enforcement)
- ✅ Fully lock-free (CAS for slot allocation)
- ✅ Realistic overhead: <1% of I/O latency in typical workloads

**When to Choose**:
- Default choice for production deployments
- Memory-constrained environments
- Typical write workloads (<100 IOPS during resync)

---

#### **Option 2: Pre-Allocated Direct Indexing**

Pre-allocate tracking for every possible chunk on disk, use chunk index for O(1) access.

```cpp
class RegionTracker {
    std::vector<std::atomic_uint32_t> _write_refcount;   // Per-chunk reference counter
    std::vector<std::atomic_uint64_t> _resync_in_flight; // Bit vector
    
    // Size: volume_size / chunk_size
    // Example: 2 TB / 32 KiB = 64M entries
    
    void set_write_in_flight(uint64_t addr, uint32_t len) {
        uint64_t start_chunk = addr / k_chunk_size;
        uint64_t end_chunk = (addr + len - 1) / k_chunk_size;
        for (uint64_t chunk = start_chunk; chunk <= end_chunk; ++chunk) {
            _write_refcount[chunk].fetch_add(1, memory_order_release);
        }
    }
    
    bool has_conflict(uint64_t addr, uint32_t len) {
        uint64_t start_chunk = addr / k_chunk_size;
        uint64_t end_chunk = (addr + len - 1) / k_chunk_size;
        for (uint64_t chunk = start_chunk; chunk <= end_chunk; ++chunk) {
            if (_write_refcount[chunk].load(memory_order_acquire) > 0) return true;
        }
        return false;
    }
};
```

**Memory Footprint**:
- 2 TB volume: **256 MB** (64M chunks × 4 bytes)
- 10 TB volume: **1.25 GB**
- Scales with volume size, not queue depth

**Performance Characteristics**:

| Operation | Latency | Notes |
|-----------|---------|-------|
| `set_write_in_flight()` | ~20-50 ns | Direct array access, 1 atomic per chunk |
| `has_conflict()` (resync) | ~20-50 ns | Direct array access, typically 1-2 chunks |
| `clear_write_in_flight()` | ~20-50 ns | Direct array access |

**Benefits**:
- ✅ **O(1) conflict detection**: Constant time regardless of concurrent write count
- ✅ No linear search overhead
- ✅ Predictable worst-case latency
- ✅ Consistent with existing Bitmap design (commit 6e4cd9b)

**Drawbacks**:
- ❌ **High memory overhead**: 256 MB per 2 TB volume
- ❌ **Wasted memory**: Most chunks never have concurrent I/O (>99.9% unused)
- ❌ **Poor cache locality**: Random chunk accesses thrash CPU cache

**When to Choose**:
- Ultra-high write workloads (>1000 IOPS during resync)
- Memory is abundant (>64 GB RAM)
- Consistency with Bitmap design philosophy is critical

---

### Recommendation

**Use Option 1 (Lock-Free Bounded Array)** as the default implementation:
- Memory efficiency matters more than 0.8% latency overhead
- Typical workloads have <100 concurrent writes, making linear search negligible
- Simpler to reason about (bounded by queue depth, not volume size)

**Consider Option 2** only if profiling shows Option 1's linear search becomes a bottleneck (unlikely unless write IOPS >> resync bandwidth).

---

## Modified Resync Flow

1. Register resync operation BEFORE reading
2. **Check for concurrent writes** (Option 1: O(n) search, Option 2: O(1) lookup)
3. If conflict detected → skip region and retry later
4. Otherwise → READ from clean, WRITE to degraded
5. Clear resync tracking and bitmap

## Modified Write Flow

1. **Register write operation** BEFORE any I/O (Option 1: find empty slot, Option 2: increment ref-count)
2. Perform write to active device
3. If degraded or write failed → dirty bitmap
4. Replicate to replica device
5. **Clear write tracking** (Option 1: clear slot, Option 2: decrement ref-count)

## Race Condition Handling

**Critical race**: Write arrives during resync read-write window

**Mitigation**: Write path marks bitmap dirty when resync is in-flight
- Ensures region gets re-copied in next sweep
- Memory ordering: acquire/release semantics on all atomics

**Reference counting** (Option 2) or **slot deduplication** (Option 1): Handles overlapping writes to same region correctly
- Multiple writes to same region → all must complete before resync sees clear

## Performance Analysis

**Write Path Overhead**:

| Approach | Per-Write Overhead | Notes |
|----------|-------------------|-------|
| Current (global) | ~20 ns | 2 atomics total |
| Option 1 (bounded array) | ~100-400 ns | Find slot (~50 ns) + slot ops |
| Option 2 (direct index) | ~320 ns | 16 chunks × 20 ns for 512 KiB write |

All options are **<0.1%** of typical NVMe write latency (500 μs for 512 KiB).

**Expected Speedup** (both options):
- Workload: 2 TB volume, random 4 KiB writes at 1000 IOPS during resync
- Current: ~50 MB/s effective resync bandwidth (90% paused)
- Proposed: ~500 MB/s effective resync bandwidth (1% paused)
- **10x speedup** (conservative estimate)

### New Files
- `src/raid/raid1/region_tracker.hpp` - Lock-free tracking (Option 1 or 2)
- `src/raid/raid1/tests/region_tracker/test_region_tracker.cpp` - Unit tests
- `src/raid/raid1/tests/concurrency/test_region_tracking.cpp` - Concurrency tests

### Modified Files
- `src/raid/raid1/raid1_resync_task.hpp` - Add RegionTracker member
- `src/raid/raid1/raid1_resync_task.cpp` - Modify `__copy_region()` and `__run()`
- `src/raid/raid1/raid1.cpp` - Modify `__replicate()` to track write regions
- `src/raid/raid1/raid1_impl.hpp` - Add RegionTracker member

### Testing Strategy

**Unit Tests**:
- RegionTracker set/clear/check operations
- Overlapping writes to same region
- Slot exhaustion handling (Option 1)
- Memory ordering (ThreadSanitizer)

**Integration Tests**:
- Write arrives during resync read-write window
- Multiple concurrent writes to same region
- Resync skip and retry logic
- Async write completion tracking

**Stress Tests**:
- 1000 random writes during resync (checksum validation)
- Full volume resync under heavy write load (performance validation)
- Queue depth saturation (128 concurrent writes)
- Degradation during resync with overlapping writes

**Performance Benchmarks**:
- Compare Option 1 vs Option 2 under varying write IOPS (10, 100, 1000)
- Measure resync throughput degradation vs current baseline
- Profile conflict detection latency distribution

## Estimated Effort

**Time**: 2-3 weeks
**Code**: ~1,500 LOC (Option 1), ~1,750 LOC (Option 2)
**Risk Level**: Medium-High (lock-free correctness, race window handling)

## Recommendation

**Implement Option 1** as the initial version:
- Simpler implementation (~250 LOC less)
- Dramatically lower memory footprint (1.5 KiB vs 256 MB)
- Negligible latency penalty in realistic workloads (<1% overhead)
- Can always migrate to Option 2 later if profiling shows bottleneck

**Only implement Option 2 if**:
- Profiling shows >5% resync overhead from conflict detection
- Write IOPS during resync consistently exceeds 1000 IOPS
- Memory is not a concern (>64 GB RAM)

For production deployments with large volumes and concurrent write workloads, the performance improvement (10-50x faster resync) justifies the implementation complexity regardless of which option is chosen.

Ilyin-Ilya:
## Problem

The previous resync mechanism used a global \`PAUSE\` state: any in-flight write — to any LBA — paused all resync I/O until every outstanding write completed. This was correct but overly conservative: a write to LBA 512 MiB had no business blocking resync of a dirty region at LBA 0.

\`sisl::atomic_status_counter\` packed the write counter and resync state into a single 64-bit word to make the counter-drains-to-zero → PAUSE→ACTIVE transition atomic. The complexity was justified by a real race, but the global-pause design remained the bottleneck.

## Solution

Replace the global PAUSE with a lock-free per-region \`RegionTracker\`. Each in-flight write registers its \`(lba, len)\` range on entry and deregisters it on completion. Resync checks for overlap **before** and **after** each chunk copy — skipping only the chunks that actually conflict. Unrelated chunks in the same dirty run proceed without any yield.

### RegionTracker (\`region_tracker.hpp\`)

Bounded flat slot array (default 256 slots — one per queue entry). Each slot is a single cache-line-aligned \`atomic<uint64_t>\` packing \`chunk_idx\` (upper 32 bits) and \`chunk_count\` (lower 32 bits), with \`UINT64_MAX\` as the free sentinel. \`track()\` claims a free slot with a single release-CAS from \`k_free\` to the packed value; \`untrack()\` frees it with a single release-CAS back to \`k_free\`. There is no transitional window — a slot is either fully occupied or fully free — so \`overlaps()\` needs no special-casing.

A **shadow completion ring** (\`_shadow\`, 4× slot count) records recently completed writes for Phase 2 detection. \`snapshot_gen()\` captures the completion generation before Phase 1; \`completed_since(gen)\` scans completions that arrived during the READ window. Multi-producer safety is achieved via \`fetch_add(acq_rel)\` to claim a distinct shadow index per caller, with a \`seq\` field as the publish barrier so readers never see partial entries.

### Two-phase conflict check in \`__run()\`

**Phase 1** (pre-copy): if \`_region_tracker.overlaps(lba, chunk)\` then advance within the dirty run and continue — do not copy, do not spend \`copies_left\` budget.

**Phase 2** (post-copy): after \`__copy_region\` succeeds, skip \`clean_region\` if either:
- \`_region_tracker.overlaps()\` is still true (write is still in-flight), **or**
- \`_region_tracker.completed_since(lba, len, gen_before)\` is true — a write arrived **and fully completed** during the READ window (slot freed before Phase 2 ran); the shadow log records this completion so the copied data is treated as potentially stale.

### Other changes

- \`PAUSE = 3\` state and \`__pause()\` removed; \`STOPPING\` renumbered from 4 → 3.
- \`sisl::atomic_status_counter\` dependency dropped; replaced by a plain \`std::atomic<resync_state>\`.
- \`ResyncWriteGuard\` updated to carry and pass \`(lba, len)\`.
- \`copies_left\` budget is only consumed by actual copy attempts (read+write to replica); Phase 1 skips are free.
- \`next_dirty()\` is called only when the current dirty run is exhausted (\`sz == 0\`), not at the bottom of every iteration.
- \`resync_skip_from\` cursor + \`next_dirty_after()\`: when an entire dirty run is blocked by in-flight writes (\`!any_copy\`), skip past it on the next sweep to prevent lower-LBA starvation.

## Tests

**\`tests/region_tracker/region_tracker_test.cpp\`** — 11 unit tests covering empty state, exact/partial/contained overlaps, adjacent-no-overlap, unregister clears slot, duplicate LBA gets two slots, fill-and-drain, concurrent register/unregister (TSAN target), plus 7 tests for \`completed_since()\`: before-snapshot not detected, overlapping detected, non-overlapping not detected, partial overlap, multiple writes with gap, shadow overflow conservative return, and a concurrent-untrack/completed_since no-false-negatives TSAN test (8 threads × 500 iterations).

**\`tests/concurrency/write_resync_no_pause.cpp\`** — 4 integration tests using \`Raid1ResyncTask\` + \`MirrorDevice\` + \`Bitmap\` directly:
- \`UnrelatedWriteDoesNotBlockResync\`: write held in-flight at 512 MiB; resync of LBA 0 must complete.
- \`ConflictingWriteSkipped_ResyncRetries\`: write held at LBA 0; resync must not read for 300 ms; must clean up after dequeue.
- \`Phase2ConflictDetectedAfterCopy\`: write registered after Phase 1 passes (while READ is in-flight); Phase 2 must keep the bitmap dirty; must clean up after dequeue.
- \`Phase2CompletedWriteDetected\`: write completes entirely during the READ window (slot freed before Phase 2); shadow log must detect it and keep bitmap dirty; verifies the \`completed_since()\` path specifically.

**\`tests/concurrency/concurrent_enqueue_dequeue.cpp\`** — concurrent enqueue/dequeue + resync stress test; TSAN is the correctness signal for data races in \`RegionTracker\`.

All existing asyncio bitmap tests pass unchanged.

### DIFF
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -98,11 +98,7 @@ Raid1Disk::Raid1Disk(boost::uuids::uuid const& uuid, std::shared_ptr< ublk_disk 
     // Initialize resync_task; slot_count = 2×qdepth so the tracker can hold every
     // in-flight write at peak depth. Fall back to 256 (= 2×128 default) when the
     // ublkpp_tgt option group is not loaded (unit test context).
-    uint32_t const resync_slots = []() -> uint32_t {
-        try {
-            return 2u * SISL_OPTIONS["qdepth"].as< uint16_t >();
-        } catch (...) { return 256u; }
-    }();
+    uint32_t const resync_slots = SISL_OPTIONS.count("qdepth") ? 2u * SISL_OPTIONS["qdepth"].as< uint16_t >() : 256u;
     _resync_task =
         std::make_shared< Raid1ResyncTask >(_dirty_bitmap, _reserved_size, block_size(),
                                             params()->basic.max_sectors << SECTOR_SHIFT, resync_slots, _raid_metrics);
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -97,7 +97,8 @@ Raid1Disk::Raid1Disk(boost::uuids::uuid const& uuid, std::shared_ptr< ublk_disk 
 
     // Initialize resync_task
     _resync_task = std::make_shared< Raid1ResyncTask >(_dirty_bitmap, _reserved_size, block_size(),
-                                                       params()->basic.max_sectors << SECTOR_SHIFT, 256, _raid_metrics);
+                                                       params()->basic.max_sectors << SECTOR_SHIFT,
+                                                       2 * SISL_OPTIONS["qdepth"].as< uint16_t >(), _raid_metrics);
 
     // Write the up-to-date superblocks and mark devices as in use
     __become_active();
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -718,8 +718,9 @@ disk_task< int > Raid1Disk::async_iov(ublksrv_queue const* q, ublk_io_data const
     auto const state = __capture_route_state();
     auto const is_discard = (op == UBLK_IO_OP_DISCARD || op == UBLK_IO_OP_WRITE_ZEROES);
 
-    // Halt the Resync Task for the duration of this write (both active and backup legs).
-    auto _guard = raid1::ResyncWriteGuard{*_resync_task};
+    // Register this write's LBA range in the region tracker so resync skips only the
+    // conflicting chunk rather than pausing globally.
+    auto _guard = raid1::ResyncWriteGuard{*_resync_task, addr, len};
     auto const bm = __compute_backup_mode(state, addr, len, is_discard);
 
     auto const adj_addr = addr + _reserved_size;
@@ -797,8 +798,9 @@ io_result Raid1Disk::sync_iov(uint8_t op, iovec* iovecs, uint32_t nr_vecs, off_t
     // WRITE / DISCARD / WRITE_ZEROES: flat replication -- mirrors async_iov write path.
     auto const is_discard = (op == UBLK_IO_OP_DISCARD || op == UBLK_IO_OP_WRITE_ZEROES);
 
-    // Halt the Resync Task before checking bitmap
-    auto _guard = raid1::ResyncWriteGuard{*_resync_task};
+    // Register this write's LBA range in the region tracker so resync skips only the
+    // conflicting chunk rather than pausing globally.
+    auto _guard = raid1::ResyncWriteGuard{*_resync_task, static_cast< uint64_t >(addr), len};
     auto const bm = __compute_backup_mode(state, static_cast< uint64_t >(addr), len, is_discard);
 
     auto const active_res = state.active_dev->disk->sync_iov(op, iovecs, nr_vecs, adj_addr);
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -99,9 +99,9 @@ Raid1Disk::Raid1Disk(boost::uuids::uuid const& uuid, std::shared_ptr< ublk_disk 
     // in-flight write at peak depth. Fall back to 256 (= 2×128 default) when the
     // ublkpp_tgt option group is not loaded (unit test context).
     uint32_t const resync_slots = SISL_OPTIONS.count("qdepth") ? 2u * SISL_OPTIONS["qdepth"].as< uint16_t >() : 256u;
-    _resync_task =
-        std::make_shared< Raid1ResyncTask >(_dirty_bitmap, _reserved_size, block_size(),
-                                            params()->basic.max_sectors << SECTOR_SHIFT, resync_slots, _raid_metrics);
+    _resync_task = std::make_shared< Raid1ResyncTask >(_dirty_bitmap, _reserved_size, block_size(),
+                                                       params()->basic.max_sectors << SECTOR_SHIFT, resync_slots,
+                                                       be32toh(_sb->fields.bitmap.chunk_size), _raid_metrics);
 
     // Write the up-to-date superblocks and mark devices as in use
     __become_active();
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -97,7 +97,7 @@ Raid1Disk::Raid1Disk(boost::uuids::uuid const& uuid, std::shared_ptr< ublk_disk 
 
     // Initialize resync_task
     _resync_task = std::make_shared< Raid1ResyncTask >(_dirty_bitmap, _reserved_size, block_size(),
-                                                       params()->basic.max_sectors << SECTOR_SHIFT, _raid_metrics);
+                                                       params()->basic.max_sectors << SECTOR_SHIFT, 256, _raid_metrics);
 
     // Write the up-to-date superblocks and mark devices as in use
     __become_active();
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -95,10 +95,17 @@ Raid1Disk::Raid1Disk(boost::uuids::uuid const& uuid, std::shared_ptr< ublk_disk 
     // Initialize bitmap and handle initial degradation based on route determination
     __init_bitmap_and_degraded_route();
 
-    // Initialize resync_task
-    _resync_task = std::make_shared< Raid1ResyncTask >(_dirty_bitmap, _reserved_size, block_size(),
-                                                       params()->basic.max_sectors << SECTOR_SHIFT,
-                                                       2 * SISL_OPTIONS["qdepth"].as< uint16_t >(), _raid_metrics);
+    // Initialize resync_task; slot_count = 2×qdepth so the tracker can hold every
+    // in-flight write at peak depth. Fall back to 256 (= 2×128 default) when the
+    // ublkpp_tgt option group is not loaded (unit test context).
+    uint32_t const resync_slots = []() -> uint32_t {
+        try {
+            return 2u * SISL_OPTIONS["qdepth"].as< uint16_t >();
+        } catch (...) { return 256u; }
+    }();
+    _resync_task =
+        std::make_shared< Raid1ResyncTask >(_dirty_bitmap, _reserved_size, block_size(),
+                                            params()->basic.max_sectors << SECTOR_SHIFT, resync_slots, _raid_metrics);
 
     // Write the up-to-date superblocks and mark devices as in use
     __become_active();
--- a/src/raid/raid1/region_tracker.hpp
+++ b/src/raid/raid1/region_tracker.hpp
@@ -72,8 +72,10 @@ public:
     // len-then-CAS sequence cannot accidentally free a slot belonging to a different write.
     void untrack(uint64_t lba, uint32_t len) noexcept {
         for (auto& slot : _slots) {
-            if (slot.lba.load(std::memory_order_relaxed) != lba) continue;
-            if (slot.len.load(std::memory_order_relaxed) != len) continue;
+            // Acquire on lba synchronizes with the release-CAS in track() so the subsequent
+            // len load is ordered correctly on weakly-ordered hardware (arm64).
+            if (slot.lba.load(std::memory_order_acquire) != lba) continue;
+            if (slot.len.load(std::memory_order_acquire) != len) continue;
             uint64_t expected = lba;
 
             // Atomically claim a shadow slot. Each concurrent caller gets a distinct gen,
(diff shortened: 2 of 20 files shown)


src/raid/raid1/region_tracker.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+    // Shadow completion log: records recently completed writes for Phase 2 detection.
+    // Not cache-line padded — only accessed by the resync reader (completed_since) and
+    // I/O completion threads (untrack). seq provides the synchronization point: producers
+    // write lba/len then store seq(release); the reader acquires on seq before reading
+    // lba/len, so no additional padding is needed between the fields.
+    //
+    // Sized to 4× the main slot count so the ring overflows only when more than
+    // 4×qdepth writes complete during a single __copy_region() call.
+    struct ShadowEntry {
```
szmyd, line 184: `ShadowEntry` is ~24 bytes, so ~2-3 entries share a cache line. With concurrent `untrack()` calls writing to adjacent shadow entries under high queue depth, this could cause false sharing between I/O completion threads. The `Slot` struct is padded to avoid exactly this for the main slot array. Was the trade-off here (memory vs. false sharing) intentional? Happy to accept it if so, just want to confirm it was a conscious choice.

src/raid/raid1/raid1.cpp
```diff
@@ -95,9 +95,17 @@ Raid1Disk::Raid1Disk(boost::uuids::uuid const& uuid, std::shared_ptr< ublk_disk
...
-    // Initialize resync_task
-    _resync_task = std::make_shared< Raid1ResyncTask >(_dirty_bitmap, _reserved_size, block_size(),
-                                                       params()->basic.max_sectors << SECTOR_SHIFT, _raid_metrics);
+    // Initialize resync_task; slot_count = 2×qdepth so the tracker can hold every
+    // in-flight write at peak depth. Fall back to 256 (= 2×128 default) when the
+    // ublkpp_tgt option group is not loaded (unit test context).
+    uint32_t const resync_slots = []() -> uint32_t {
+        try {
+            return 2u * SISL_OPTIONS["qdepth"].as< uint16_t >();
+        } catch (...) { return 256u; }
```
szmyd: `catch (...)` swallows all exceptions here, not just "key not registered" — a type mismatch on `.as<uint16_t>()` would silently fall back to 256 rather than surfacing the bug. `SISL_OPTIONS.count()` is the idiomatic `sisl::options` way to test for presence:
```cpp
uint32_t const resync_slots =
    SISL_OPTIONS.count("qdepth") ? 2u * SISL_OPTIONS["qdepth"].as< uint16_t >() : 256u;
```

szmyd (approved):
Looks good; will leave it up to you if you want to address the two comments.

szmyd (changes requested):
Need version bump and changelog entry. Maybe some docs/README updates too.

src/raid/raid1/region_tracker.hpp
Ilyin-Ilya: Yes, intentional — the trade-off is documented in the `ShadowEntry` comment block. The shadow ring is accessed by only two parties: `untrack()` (I/O completion threads writing) and `completed_since()` (resync thread reading). The `seq` field already serializes the producer→consumer handoff: the reader acquires on `seq` before touching `lba`/`len`, so there is no torn-read risk between producers and the single reader. The remaining concern is write–write false sharing between concurrent `untrack()` callers writing to adjacent entries. At default qdepth=128 the shadow ring is 512 entries × 24 bytes ≈ 12 KiB — small enough to fit entirely in L1. False sharing still causes write-invalidation traffic between cores, but the ring being hot in L1 bounds the impact. Padding each entry to 64 bytes would triple the ring to ~36 KiB and push it into L2 on most microarchitectures, trading cache-miss latency on every `completed_since()` scan for lower write-invalidation traffic between `untrack()` callers. The scan path seemed more critical to keep fast. Happy to revisit if `untrack()` ever shows up hot in perf.

src/raid/raid1/raid1.cpp
Ilyin-Ilya: Good catch — fixed. Replaced the `try/catch(...)` with `SISL_OPTIONS.count("qdepth")` so a type mismatch on `.as<uint16_t>()` surfaces as an exception rather than silently falling back to 256. Pushed in eb397c7.

szmyd (approved):
## Good luck!
Image

PR #256, opened 2026-05-14
refactor(cqe_state): build on sisl::async::cqe_awaitable

szmyd:
Replace ublkpp's standalone cqe_state awaitable with a thin derivative of sisl::async::cqe_awaitable. The new cqe_state adds only _owner (for ublksrv error-path tagging); all awaitable fields and protocol (_result, _result_ready, _waiter, await_ready/suspend/resume) are inherited.

Dispatch loops now call sisl::async::complete_cqe_state(*state, res) instead of manually writing fields and resuming _waiter, removing the try/catch that was guarding h.resume() in run_queue_loop and mock_ublksrv. The managed-bit helpers (is_managed_user_data, decode_managed_user_data, encode_managed_user_data) replace the raw k_target_bit bit-ops at all decode sites; k_target_bit is kept as a backward-compat alias.

Also: mark async_iov noexcept throughout the stack (base class + all overrides
+ test mock) and change disk_task::promise_type::unhandled_exception to call std::terminate instead of rethrowing. Any uncaught exception in a coroutine body would propagate through the noexcept _on_complete callback anyway; making the contract explicit here and letting ublksrv USER_RECOVERY restart the process is safer than silently returning -EIO for one tag.

### DIFF
--- a/include/ublkpp/lib/cqe_state.hpp
+++ b/include/ublkpp/lib/cqe_state.hpp
@@ -1,11 +1,11 @@
 #pragma once
 
-#include <coroutine>
 #include <cstdint>
 #include <utility>
 #include <vector>
 
 #include <liburing.h>
+#include <sisl/async/cqe_state.hpp>
 #include <sisl/logging/logging.h>
 #include <ublksrv.h>
 
@@ -27,9 +27,11 @@ namespace ublkpp {
 //   }
 //
 // The queue's CQE loop (run_queue_loop in src/target/ublkpp_tgt.cpp) inspects
-// bit 63 (k_target_bit) of cqe->user_data: if set, the remaining bits are a
-// cqe_state*; the loop writes _result, marks _result_ready, and resumes _waiter.
-// If clear, the CQE is delegated to ublksrv as a command completion.
+// sisl::async::is_managed_user_data(cqe->user_data): if set, the remaining bits
+// are a cqe_state*; the loop calls sisl::async::complete_cqe_state(*state, res)
+// which writes _result, marks _result_ready, and resumes _waiter via the
+// callback installed by await_suspend. If clear, the CQE is delegated to
+// ublksrv as a command completion.
 //
 // Two flavors of cqe_state:
 //   - Per-IO (build_cqe_state_data path): allocated in async_io::_pool, owned
@@ -37,22 +39,24 @@ namespace ublkpp {
 //     with -EIO if the coroutine throws.
 //   - Stand-alone service loops: coroutine-frame-local, _owner = nullptr, and
 //     callers handle their own errors. Encode the user_data manually with
-//     `reinterpret_cast<uint64_t>(state) | k_target_bit`.
+//     sisl::async::encode_managed_user_data(state).
 //
 // Reference implementation: src/driver/fs_disk.cpp.
 // =============================================================================
 
-// Bit 63 of cqe->user_data distinguishes target SQEs (custom-driver async I/O)
-// from ublksrv command SQEs (FETCH/COMMIT). On x86_64/ARM64, canonical userspace
-// addresses use <=48 bits, so bit 63 is always zero in any valid pointer; the OR
-// is safe and reversible with `& ~k_target_bit`.
-//
-// Probe timeout CQEs reuse k_target_bit with a null pointer (user_data == k_target_bit).
-// run_queue_loop checks state == nullptr after stripping the bit to distinguish them from
-// real I/O CQEs.
-constexpr uint64_t k_target_bit = 1ULL << 63;
+struct async_io;
 
-struct cqe_state;
+// Extends sisl::async::cqe_awaitable with _owner for ublksrv error reporting.
+// Inherits: _result, _result_ready, _waiter, _on_complete, _on_complete_ctx
+//           await_ready(), await_suspend(), await_resume()
+// Use sisl::async::complete_cqe_state(*state, res) to deliver a CQE result.
+//
+// _owner is nullable: per-IO cqe_states (build_cqe_state_data path) point at the slot's
+// async_io so errors can be reported via ublksrv_complete_io. Stand-alone cqe_states set
+// _owner = nullptr; callers handle their own errors.
+struct cqe_state : sisl::async::cqe_awaitable {
+    async_io* _owner{nullptr};
+};
 
 // Per-IO state tracking one inflight request, owned by the ublksrv-allocated io_data slot.
 //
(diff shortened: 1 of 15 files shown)


include/ublkpp/lib/disk_task.hpp
```diff
@@ -43,7 +43,7 @@ struct disk_task {
         final_awaiter final_suspend() noexcept { return {}; }
 
         void return_value(T v) noexcept { _value = v; }
-        [[noreturn]] void unhandled_exception() { throw; }
+        [[noreturn]] void unhandled_exception() noexcept { std::terminate(); }
```
Ilyin-Ilya, line 46: The switch from `throw;` to `std::terminate()` is a significant severity escalation: previously one failing I/O returned -EIO to the block layer; now any unhandled exception terminates the process (recovered by USER_RECOVERY). This is internally consistent because `async_iov` is now `noexcept` and the old try/catch in `run_queue_loop` is gone, but the rationale is easy to miss for a future reader.

Suggested: add a brief comment explaining the intent:

```cpp
// Intentional: async_iov is noexcept, so exceptions should never reach here.
// If one somehow does (broken override), terminating is safer than silently
// corrupting state. USER_RECOVERY restarts the process automatically.
[[noreturn]] void unhandled_exception() noexcept { std::terminate(); }
```

src/lib/tests/test_cqe_state.cpp
```diff
@@ -109,42 +107,26 @@ TEST(BuildCqeStateData, ReturnedPointerMatchesDecodedUserData) {
...
+    sisl::async::complete_cqe_state(state, 7); // writes _result, _result_ready, resumes _waiter
     EXPECT_EQ(captured, 7);
 }
 
 TEST(cqe_state, FastPathSkipsSuspendWhenAlreadyReady) {
     ublkpp::async_io io{};
-    ublkpp::cqe_state state{._owner = &io, ._result = 55, ._result_ready = true};
+    ublkpp::cqe_state state{};
+    state._owner = &io;
+    state._result = 55;
```
Ilyin-Ilya, line 128: The `FastPathSkipsSuspendWhenAlreadyReady` test directly writes `state._result = 55` and `state._result_ready = true`, which are inherited fields from `sisl::async::cqe_awaitable`. This couples the test to sisl's internal layout — if sisl renames or restructures those fields the test silently becomes incorrect rather than failing to compile.

The public API for pre-completing a `cqe_state` is `sisl::async::complete_cqe_state(state, 55)`. When no waiter is installed, `complete_cqe_state` should mark the result ready without resuming anything, so `await_ready()` returns `true` on the subsequent `co_await`. Prefer:

```cpp
sisl::async::complete_cqe_state(state, 55); // marks ready; no waiter to resume
int captured = -1;
await_cqe_and_capture(&state, &captured); // await_ready=true -> no suspension
```

If `complete_cqe_state` doesn't support the no-waiter path, that is worth a separate issue — but the current code would mask it.

src/lib/tests/test_cqe_state.cpp
```diff
@@ -109,42 +107,26 @@ TEST(BuildCqeStateData, ReturnedPointerMatchesDecodedUserData) {
 // Coroutine that co_awaits *state and writes the result to *out.
```
Ilyin-Ilya, line 107: Three tests that exercised `await_ready()` and `await_resume()` in isolation were deleted (`AwaitReadyFalseWhenNotReady`, `AwaitReadyTrueWhenReady`, `AwaitResumeReturnsResult`). Since `cqe_state` now inherits these from `sisl::async::cqe_awaitable`, testing them is belt-and-suspenders against a future sisl regression.

Consider adding `EXPECT_TRUE(state.await_ready())` to `FastPathSkipsSuspendWhenAlreadyReady` before the `co_await`, so the fast-path is verified explicitly rather than only implicitly through the coroutine's return value.

include/ublkpp/lib/cqe_state.hpp
```diff
@@ -69,37 +73,19 @@ struct async_io {
...
-    void await_suspend(std::coroutine_handle<> h) noexcept { _waiter = h; }
-    int await_resume() const noexcept { return _result; }
-};
-
 inline cqe_state* async_io::next_state() {
     RELEASE_ASSERT_LT(_pool.size(), _pool.capacity(),
                       "cqe_state pool exhausted; prepare_result::max_sqes_per_io underestimated")
-    _pool.push_back(cqe_state{._owner = this});
+    _pool.push_back(cqe_state{});
+    _pool.back()._owner = this; // designated-init can't reach inherited members in C++20
```
Ilyin-Ilya, line 80: The comment reads:

```cpp
_pool.back()._owner = this; // designated-init can't reach inherited members in C++20
```

Citing `C++20` implies the restriction might be lifted in a later standard, but it will not be — designated initializers cannot name inherited base-class members in any version of C++ (it is part of the aggregate-initialization rules, not version-gated). Suggested wording:

```cpp
_pool.back()._owner = this; // C++ designated-init cannot reach inherited base-class members
```

include/ublkpp/lib/ublk_disk.hpp
```diff
@@ -82,7 +82,7 @@ class ublk_disk {
     // Callers (e.g. Raid0Disk) may pass a pointer to a loop-local iovec that is destroyed after
     // start() returns; any access past the first suspension point is a use-after-free.
     virtual disk_task< int > async_iov(ublksrv_queue const* /*q*/, ublk_io_data const* /*data*/, iovec* /*iovecs*/,
-                                       uint32_t /*nr_vecs*/, uint64_t /*addr*/) {
+                                       uint32_t /*nr_vecs*/, uint64_t /*addr*/) noexcept {
```
Ilyin-Ilya, line 85: `noexcept` on a coroutine factory is narrower than it appears: the guarantee only covers the initial frame-allocation call, not the body. Exceptions thrown inside the coroutine body are routed to `promise_type::unhandled_exception` regardless of `noexcept` on the factory signature. Overrides that add `noexcept` are implicitly committing that the body never throws (enforced at runtime via `std::terminate()` in `unhandled_exception`), not at compile time.

A short comment would help future implementors understand what they are signing up for:

```cpp
// noexcept covers only the coroutine frame-allocation call, not the body.
// Exceptions in the body are caught by promise_type::unhandled_exception
// and terminate the process. Override implementations must not throw.
virtual disk_task< int > async_iov(...) noexcept {
```

src/raid/raid1/raid1.cpp
```diff
@@ -344,8 +349,12 @@ bool Raid1Disk::__swap_device(std::string const& outgoing_device_id, std::shared
     }
     // Commit SuperBlock to new device; if this fails it's not fatal per say...could work
```
Ilyin-Ilya, line 350: Two issues in this comment:

1. **Typo**: `per say` should be `per se`.
2. **Stale framing**: The comment was written before error-checking was added. Now that a `RLOGW` is logged on failure, the "optimistic" framing is still accurate but the old wording reads as if the failure is silently swallowed.

Suggested:

```cpp
// Commit SuperBlock to new device; if this fails it's not fatal per se — we'll
// retry when we become clean — so log the warning and proceed optimistically.
```

src/tests/mock_ublksrv/mock_ublksrv.cpp
```diff
@@ -106,19 +106,16 @@ io_result MockUblksrv::submit_io(int tag, uint8_t op, uint64_t start_sector, uin
...
-    if (!(cqe->user_data & k_target_bit)) return (void)io_uring_cqe_seen(&_ring, cqe);
-    auto* state = reinterpret_cast< cqe_state* >(cqe->user_data & ~k_target_bit);
+    if (!sisl::async::is_managed_user_data(cqe->user_data)) return (void)io_uring_cqe_seen(&_ring, cqe);
+    auto* state = static_cast< cqe_state* >(sisl::async::decode_managed_user_data(cqe->user_data));
     if (!state || !state->_owner) return (void)io_uring_cqe_seen(&_ring, cqe);
     int const tag = state->_owner->_tag;
     int const res = cqe->res;
 
     // Consume the CQE immediately so peek sees the next one
     io_uring_cqe_seen(&_ring, cqe);
```
Ilyin-Ilya, line 117: The comment above `io_uring_cqe_seen` explains why the CQE is consumed before completion. A note that `complete_cqe_state` resumes the coroutine synchronously (same thread, same call stack, before returning to the caller) would clarify why the call ordering here is safe — in particular why the completion bookkeeping on the following line is valid:

```cpp
// Consume the CQE immediately so peek sees the next one.
io_uring_cqe_seen(&_ring, cqe);
// complete_cqe_state resumes the coroutine synchronously on this thread.
sisl::async::complete_cqe_state(*state, res);
```

Ilyin-Ilya (commented):
A few genuine findings from the automated review pass below. False positives (A2-1/A2-2 about `run_resync_queue_loop` and resync thread, A3-1/A3-2/A3-3 about `__become_clean` return type, `noexcept` on `async_iov`, and destructor error-checking) are all already addressed in the diff and discarded.

PR #261, opened 2026-05-20
feat(raid1): SB v2 — fixed bitmap disk layout + tighter user-data alignment

Ilyin-Ilya:
## Summary

- **Fixed on-disk reserved region (SB v2)**: \`_reserved_size\` is now always
  \`sizeof(SuperBlock) + k_superbitmap_bits × k_page_size\` (~125.6 MiB) regardless of capacity,
  replacing the old variable formula proportional to disk size. This reserves headroom for future
  volume resize without a format change.
- **\`init_to\` writes only \`_num_pages\` (capacity-derived)**: the reserved region is claimed by
  \`_reserved_size\`, not by pre-writing zeros. Tail pages beyond \`_num_pages\` are unreachable
  because \`load_from\` gates every read on \`_super_bitmap.test_bit(pg_idx)\`, which is only set by
  \`sync_to\` after a successful write. Garbage in the unrewritten tail is never loaded.
- **Tighter user-data alignment (SB v2)**: \`_reserved_size\` padding now aligns to \`logical_bs\`
  (~4 KiB) instead of \`max_sectors_bytes\` (~512 KiB), reclaiming up to ~511 KiB of wasted tail
  space per device. v1 arrays keep the old alignment exactly.
- **\`SB_VERSION\` bumped 1 → 2**: signals both format changes above. \`load_superblock\` now
  **rejects** existing disks with version < 2 (returns \`std::errc::invalid_argument\`) — there is
  no safe in-place migration since the user-data offset moves.
- **Constructor call order fixed**: \`__load_and_select_superblock\` now runs before
  \`__init_params\` so the SB version is available when choosing the alignment policy.
  \`__init_params\` signature simplified to \`void __init_params()\`.
- **All tests updated**: \`Times(N)\` combined bitmap+SB expectations replaced with addr-based
  GMock matchers (\`Times(AtLeast(1)).WillRepeatedly\` for bitmap writes at addr>0, \`Times(1)\` for
  superblock at addr==0).

## Motivation

Previously \`_reserved_size = sizeof(SB) + ceil(capacity / page_width) × k_page_size\`, so each
disk had a different reserved region size. Swapping disks or resizing volumes required careful
bookkeeping around that variable layout. The new design gives every v2 disk the same fixed
footprint, making the on-disk format fully predictable and enabling future resize without a new SB
version bump.

**Max supported disk size is unchanged**: \`k_superbitmap_bits × k_min_chunk_size × k_page_size × k_bits_in_byte ≈ 31.42 TiB\`.

**Breaking change**: v1 arrays used a smaller, capacity-proportional reserved region. Loading with
v2 code would misalign the user-data region. Old arrays must be re-created. The \`SB_VERSION = 2\`
bump and the explicit rejection in \`load_superblock\` signal this.

## Test plan

- [ ] \`Raid1.InitBitmap\` — verifies exactly \`ceil(_num_pages / max_pages)\` write calls, total bytes = \`_num_pages × k_page_size\`
- [ ] \`Raid1.InitBitmapFailure\` — first batch failure throws \`std::runtime_error\`
- [ ] \`Raid1.V1SuperblockRejected\` — v1 SB on device_a throws; \`Times(0)\` confirms device_b is never touched
- [ ] \`Raid1.NewArrayWritesV2Superblock\` — both SB writes during construction carry \`version == htobe16(2)\`
- [ ] \`Raid1.InitSuperBlock\`, \`NewDeviceA/B\`, \`ReversedOrder\`, \`NewDeviceThrowCantDirty\` — addr-split expectations verified
- [ ] \`Raid1.SwapDeviceA/B\`, \`SwapDeviceWhileIdle\` — fixed from \`Times(2)\` combined expectations
- [ ] \`Raid1.IdenticalDeviceProbing\`, \`DiffereingDeviceProbing\` — capacity assertions use \`reserved_size()\` dynamically, pass with new alignment
- [ ] All 13 test suites pass on SSH build host


src/raid/raid1/bitmap.cpp
```diff
@@ -98,12 +98,26 @@ void Bitmap::init_to(std::shared_ptr< ublk_disk > device) {
...
+    for (auto pg_idx = 0UL; _num_pages > pg_idx; ++pg_idx) {
+        if (_page_map[pg_idx].page.load(std::memory_order_relaxed)) continue;
+        void* new_mem{nullptr};
+        if (auto err = ::posix_memalign(&new_mem, _align, k_page_size); err)
+            throw std::runtime_error("OutOfMemory"); // LCOV_EXCL_LINE
+        memset(new_mem, 0, k_page_size);
+        _page_map[pg_idx]._page_mem = new_mem;
+        _page_map[pg_idx].page.store(reinterpret_cast< word_t* >(new_mem), std::memory_order_relaxed);
+        // loaded_from_disk stays false — these are new pages, not loaded from disk.
+    }
```
szmyd: Why do we pre-allocate clean bitmap pages that are likely to never be used? This will make every raid1 volume always use the _max_ memory (~128MiB) right?

src/raid/raid1/raid1_superblock.cpp
```diff
@@ -74,11 +74,11 @@ io_result write_superblock(ublk_disk& device, raid1::SuperBlock* sb, bool device
 // the superblock to the current version. Otherwise migrate any changes needed after version discovery.
 std::expected< std::pair< raid1::SuperBlock*, bool >, std::error_condition >
 load_superblock(ublk_disk& device, boost::uuids::uuid const& uuid, uint32_t const chunk_size) {
-    auto sb = read_superblock(device);
+    auto sb = std::unique_ptr< raid1::SuperBlock, decltype(&free) >(read_superblock(device), free);
```
Ilyin-Ilya, line 78: this ugly thing is because sb is allocated with memset, not new, so we have to delete it with free() 
unique_ptr helps us to avoid manually calling frees

src/raid/raid1/raid1_superblock.cpp
```diff
@@ -104,7 +104,17 @@ load_superblock(ublk_disk& device, boost::uuids::uuid const& uuid, uint32_t cons
               be32toh(sb->fields.bitmap.chunk_size), chunk_size, to_string(uuid))
     }
 
-    if (SB_VERSION > be16toh(sb->header.version)) { sb->header.version = htobe16(SB_VERSION); }
-    return std::make_pair(sb, was_new);
+    if (was_new) {
+        // Fresh device — stamp with current version.
+        sb->header.version = htobe16(SB_VERSION);
+    } else if (SB_VERSION > be16toh(sb->header.version)) {
```
szmyd: V1 arrays can be supported without a breaking change - the `_reserved_size` formula is the only incompatibility, and the version-branched alignment already in `__init_params` is the right model for this.

**Suggested approach:**

1. **`load_superblock`**: remove this rejection block; return the existing v1 SB as-is without stamping a new version.
2. **`__init_params`**: branch on `sb_version < 2` for the size formula:
```cpp
if (sb_version < 2) {
    // v1: capacity-proportional (preserves the exact on-disk layout).
    auto const bitmap_size =
        ((our_params.basic.dev_sectors << SECTOR_SHIFT) / k_min_chunk_size) / k_bits_in_byte;
    _reserved_size = sizeof(SuperBlock) + bitmap_size;
} else {
    // v2+: fixed maximum, leaves headroom for future resize.
    _reserved_size = sizeof(SuperBlock) + (k_superbitmap_bits * k_page_size);
}
```
The alignment branch below already handles the rest (v1 pads to `max_sectors_bytes`, v2 to `logical_bs`), so the full on-disk layout for v1 arrays is reconstructed from the version field alone. SB v2 as a resize invariant can be enforced separately when resize is implemented.

szmyd (changes requested):
One item worth addressing before merge:

szmyd (approved):
Looks great!

PR #262, opened 2026-05-20
feat(md): md-raid 1.2 migration onto ublkpp's RAID stack

szmyd:
## Summary

- Adds `ublkpp::md::disk_type` enum (`md_none`, `md_native`, `md_stamped`, `md_dirty`) to the public API
- Adds `ublkpp::md::probe(disk_handle)` for non-destructive superblock detection; reads ~8 KiB from the device and returns which superblock is present without writing anything
- Adds `make_md_raid1_disk` and `make_md_raid10_disk` factory functions for one-way migration of clean md 1.2 arrays onto ublkpp's RAID stack; user data is preserved byte-identical at md's `data_offset` via a piecewise address translation shim (`MdDisk`)
- `MdDisk` constructor is fully read-only: probes topology and validates UUID/cross-leg consistency with zero writes to any device. All writes (own SB at byte 0, md SB scrub at byte 4096) are deferred to the first `prepare()` call via `_stamp_pending` atomic. This ensures no device is permanently modified if any leg fails validation - matching the established RAID1 "no writes before prepare" invariant and fixing a bug where devices were stamped individually during factory construction before cross-leg validation completed
- Extracts `classify_md_sb()` helper to eliminate duplication between `probe()` and `discovered_topo_from_md()` for md SB magic + dirty-flag classification
- ublk_nublox: `add_device_open_iscsi` now probes each leg on attach and routes to `make_md_raid*_disk` or the plain RAID factories based on the detected superblock type; dirty arrays return `std::errc::device_or_resource_busy` which maps to gRPC `ABORTED` with `"md_array_dirty:"` prefix so CSI can fall back to assembling with md-raid

## Test plan

- [x] All existing md tests pass (including 3 tests updated to call `prepare()` before asserting stamp effects: `ScrubsMdSbAfterWrap`, `ReattachReadsOwnSbWithoutMdSb`, `HeadZoneShiftsByOnePage`)
- [x] New `MdProbe.NoWriteBeforePrepare` verifies construction leaves byte 0 (own SB) and byte 4096 (md SB) unmodified
- [ ] `ReattachReadsOwnSbWithoutMdSb` correctly exercises the reattach path (first wrap calls `prepare()` so the md SB is scrubbed; second wrap reads MdDisk's own SB)
- [x] Address sanitizer and thread sanitizer CI jobs pass
- [x] Manual integration test: md RAID10 array across 4 devices imports cleanly; devices are stamped only after the full stack is prepared
- [ ] Reattach: restart ublk_nublox against same devices; probe returns `md_stamped` and routes without re-importing
- [x] Dirty-array: present a dirty md array; confirm gRPC `ABORTED` with `"md_array_dirty:"` prefix and no writes to any device


Ilyin-Ilya (changes requested):
## Review: feat(md) — md-raid 1.2 migration

Five review angles (address translation, concurrency, validation, tests, API) plus a verification pass. One data-corruption blocker and several major issues must be resolved before merge.

---

### 🔴 BLOCKER — `__compute_raid1_reserved` uses the v1 formula; RAID1 stamps v2 SBs

**This causes data corruption on every freshly-imported md array.**

`MdDisk::__probe_and_init` computes `_data_band_threshold = R + md_chunk_size` where R comes from `__compute_raid1_reserved` (the v1 capacity-proportional formula, ≈8 MiB for a 2 TiB array). But the RAID1 that sits above MdDisk reads its superblock via `sync_iov(READ, addr=0)`, which MdDisk translates to **leaf byte 4096** (the md 1.2 SB location). RAID1 sees no raid1 magic there → `was_new = true` → stamps `SB_VERSION = 2` → computes `_reserved_size = sizeof(SuperBlock) + k_superbitmap_bits × k_page_size ≈ 125.69 MiB` (the v2 fixed formula already on main).

MdDisk's threshold is ≈76 KiB; RAID1's `_reserved_size` is ≈125.69 MiB. RAID1 writes bitmap pages at upper addresses 76 KiB … 125.69 MiB. MdDisk translates those as data-band writes via `+_data_band_shift`, landing them at `data_offset_md … data_offset_md + 125.6 MiB` on the leaf — i.e. **directly into user data**. This destroys up to 125 MiB of user data per leg on the first bitmap flush after mount. Verified against szmyd/ublkpp's current main (the v2 formula is already there, not waiting on any pending PR).

**Fix options:**

1. Pre-seed a valid v1 RAID1 superblock inside the head-zone before calling `make_raid1_disk`. With a v1 SB in place, `load_superblock` won't see a blank page at byte 4096, won't set `was_new`, and RAID1 will reconstruct `_reserved_size` using the capacity-proportional formula — which is exactly what `__compute_raid1_reserved` computes. This is the minimal fix.

2. Alternatively, compute `_data_band_threshold` using the v2 formula (`sizeof(SB) + k_superbitmap_bits × k_page_size ≈ 125.69 MiB`) and require `data_offset_md > 125.76 MiB`. Most real md arrays have a much smaller dead zone (typically 1 MiB), so this breaks compatibility with all real arrays and is not viable.

Option 1 is the correct path: `MdDisk` must write a minimal v1 RAID1 SB into the head zone as part of `__stamp()`, so RAID1 can read it back, recognise it, and use the v1 formula.

---

### 🔴 BLOCKER — `conanfile.py` version must be 0.33.0, not 0.32.1

The CHANGELOG correctly adds `[0.33.0]` for this PR (new public API namespace `ublkpp::md`, three new factories, one new free function, one new enum). A patch-level bump (`0.32.1`) is semantically wrong for a release that adds new public symbols. The `[0.33.0]` entry is also missing its release date.

---

### 🟠 MAJOR — `__stamp()` has no crash barrier between its two writes

`__stamp()` writes MdDisk SB at byte 0, then zeros the md SB at byte 4096. Without an `fdatasync` (or equivalent iouring flush) between them, storage can persist write 2 before write 1. After a crash in that window: byte 0 has no magic, byte 4096 is zeroed → the device appears blank to both ublkpp and md-raid. The array cannot be reattached; the user must reconstruct or restore from backup. User data is physically intact but inaccessible.

The safe sequence is: write byte 0 → `fdatasync` → write byte 4096. The first write alone (MdDisk SB without md SB scrub) is a safe intermediate state — ublkpp reattaches correctly, and an md-raid reassembly attempt on a device that has both superblocks is a user-error scenario.

---

### 🟠 MAJOR — Missing leg silently fails in the example instead of degraded mode

In `create_raid1` and `create_raid10` (example/ublkpp_disk.cpp), if only one leg path exists on disk, `probe()` returns the type of the existing leg and the code routes to `make_md_raid1_disk` / `make_md_raid10_disk`. The factory then constructs a `MdDisk` for the missing path, which calls `leaf_pread` on a non-existent file, throws `std::runtime_error`, which is caught and swallowed, and the function returns `EPERM` with no useful error message.

There is no degraded-mode import path for md arrays. Either the example must check for a missing leg and fall back gracefully (perhaps with `make_missing_disk()`), or the factory must accept a missing-leg sentinel and document the limitation. As-is, any single-disk failure makes an md array unreachable through ublkpp with a misleading error.

---

### 🟠 MAJOR — `ReattachRejectsWrongVolumeUuid` tests the wrong code path

The comment says the first `MdDisk` construction "stamps the MdDisk SB," but `prepare()` is never called. The stamp does not happen. When the second construction runs with `wrong_uuid{}`, byte 0 still has no MdDisk magic, so it takes the **fresh-import path** (reads the md SB at byte 4096), parses the md UUID, and rejects the mismatch. The reattach-path UUID guard (the `k_magic` branch in `__probe_and_init`) has zero test coverage.

Fix: call `md->prepare(nullptr, 0)` in the first block, then verify the second construction fails via the reattach guard.

---

### 🟠 MAJOR — `legacy_v1_alignment.cpp` WRITE mock returns first-iov length only

The mock lambda returns `iovecs->iov_len` (the first iov's length) rather than `iovec_len(iovecs, iovecs + nr_vecs)`. For the single-iov writes this test exercises today it is correct by coincidence, but `Bitmap::sync_to` and `init_to` batch up to `max_pages` iovecs per call. If any future test path exercises those through this mock, it will silently report a short write, masking real bugs. Use `ublkpp::iovec_len(iovecs, iovecs + nr_vecs)` as in the rest of the test suite.

---

### 🟡 MINOR — Dead `md_stamped` case in `discovered_topo_from_md`

`classify_md_sb` only returns `md_none`, `md_dirty`, or `md_native` — never `md_stamped`. The `case disk_type::md_stamped:` in the switch inside `discovered_topo_from_md` is unreachable dead code. Its presence implies the function could be called on an already-stamped leg, which it cannot (the own-SB path is handled in `__probe_and_init` before this call). Remove it or add a `[[fallthrough]]` note explaining that this case is structurally impossible.

---

### 🟡 MINOR — `far_copies == 1` accepted without justification

The layout check (`far_copies > 1`) accepts `far_copies == 1`. The inline comment says "the kernel ignores far_offset when far_copies<=1", but `far_copies == 1` is not a no-op in the kernel — it places a far copy at a strided offset. If a real near=2, far=1 array is presented, MdDisk's flat `+_data_band_shift` translation won't match the far-copy scatter, corrupting reads. If `far_copies == 1` is guaranteed never to appear in the field (mdadm's default is `far_copies=0`), the condition should be `far_copies != 0` to be explicit, or a comment should cite the invariant.

---

### 🟡 MINOR — `probe()` throws `std::runtime_error` but header documents `std::system_error`

`include/ublkpp/raid.hpp` says "Throws `std::system_error` if the device cannot be opened or read." The implementation throws `std::runtime_error` with a formatted string. A `std::system_error` would carry the actual `std::error_condition` from `io_result` and give callers a structured error code. Either match the implementation to the documentation or update the docs.

---

### 🟡 MINOR — Stale "SKIP" in `DEBUG_ASSERT` comment

`raid1.cpp`, `sync_iov`: `"backup_write must not be SKIP when become_degraded succeeds"` — `WriteBackupMode::SKIP` is deleted in this PR. The assertion logic is correct (`backup_write` must be `true`); only the message string is stale. Update to: `"backup_write must be true when become_degraded succeeds"`.

---

### 🔵 NIT — `conanfile.py` coupling: `md_disk.cpp` reimplements RAID1 sizing formula

`md_disk.cpp` includes `raid/raid1/raid1_superblock.hpp` specifically to clone RAID1's internal `_reserved_size` calculation via `__compute_raid1_reserved`. This coupling means any change to RAID1's internal formula silently breaks MdDisk's address translation — exactly the situation we're in with the v2 blocker above. The formula should live in a shared constants header that both files include, or `Raid1Disk` should expose a `reserved_size_for_capacity(uint64_t capacity_bytes)` static helper.

---

### 🔵 NIT — `disk_type` enum values have redundant `md_` prefix inside `namespace md`

`ublkpp::md::disk_type::md_native` — the `md_` prefix duplicates the namespace. Consider `ublkpp::md::disk_type::native` / `stamped` / `dirty` / `none`, or follow the project's `SCREAMING_SNAKE_CASE` convention for scoped enums.

---

**Summary:**

| Severity | Count |
|---|---|
| 🔴 Blocker | 2 |
| 🟠 Major | 4 |
| 🟡 Minor | 4 |
| 🔵 Nit | 2 |

The md-import concept is the right approach and the address-translation architecture (head-zone shift + data-band shift with a convergence loop for R) is elegant. The fix for the blocker is contained: pre-seed a v1 RAID1 SB in the head zone so RAID1 reads it back and uses the matching formula. Once that's resolved and the test gaps are addressed, this can ship.

PR #270, opened 2026-05-21
fix(raid1): remove OPTIMISTIC write path to eliminate resync stale-read race

szmyd:
## Summary

- **Root cause**: `WriteBackupMode::OPTIMISTIC` called `clean_region()` before `untrack()`, so the bitmap could be marked CLEAN while the resync task's concurrent `__copy_region` wrote stale data to the dirty mirror. Because `clean_region()` fires before the shadow log entry is published, Phase 2's `completed_since()` fires too late — there is an irreducible window where the bitmap is clean but the dirty mirror holds stale data, and reads can be routed to it.
- **Fix**: Remove `OPTIMISTIC` entirely. `clean_region()` must only be called from within the resync task's own `__run()` loop — calling it from the I/O path is inherently unsafe because the resync always holds a snapshot that may predate the caller's write. Dirty-region writes in degraded mode now always use `SKIP` (active mirror only); the resync task remains the sole authority on when a region is clean.
- **Cleanup**: `WriteBackupMode::OPTIMISTIC` removed from the enum; `__compute_backup_mode` and its now-dead `is_discard` parameter simplified away; `write_optimistic.cpp` deleted.

## Root Cause Detail

```
OPTIMISTIC:  ... write D2 to dirty → clean_region() → co_return → untrack()
                                           ↑ bitmap CLEAN
                                                           reads can land here
                                                             ↓
Resync copy:  read D1 ───────────────────────────── write D1 ────────── dirty
                                                                          ↓
Phase 2:                                              completed_since()=true
                                                      (too late; bitmap already CLEAN)
```

## Test plan
- [x] Existing test suite passes (all concurrency, syncio, bitmap, superblock tests)


Ilyin-Ilya (dismissed):
Clean, well-scoped removal of a race-prone optimization. The diff correctly eliminates `WriteBackupMode::OPTIMISTIC`, its computation function, the public `clean_region()` call site, and associated tests. Housekeeping (CHANGELOG, version bump) is present. Two real issues require attention before merge, and two documentation references to the removed mode were missed.

---

## Request Changes

### Stale documentation: `README.md` line 107

`README.md` line 107 still reads:
```
- Optimistic write recovery with dirty-bitmap-guided resync
```
This feature has been removed by this PR. The line should be deleted or replaced with an accurate description of the current resync behavior (e.g., "Background resync with region-level write coordination").

---

### Stale documentation: `docs/error_codes.md`

`docs/error_codes.md` contains a table row that documents the now-deleted `OPTIMISTIC` write mode:
```
| WRITE | Active and backup both succeed (OPTIMISTIC mode) | active result | - | Chunk cleaned in bitmap; backup marked available |
```
This row must be removed or rewritten to reflect the current two-mode behavior (`backup_write=true` / `backup_write=false`).

---

## Comments (non-blocking)

### DEBUG_ASSERT message wording — `raid1.cpp` (`async_iov` and `sync_iov`)

Both assert sites were updated to:
```cpp
DEBUG_ASSERT(backup_write, "backup_write must not be SKIP when become_degraded succeeds");
```
`backup_write` is now a `bool`, not an enum, so "must not be SKIP" is misleading — it will appear verbatim in crash logs. Suggested wording:
```cpp
DEBUG_ASSERT(backup_write, "backup_write must be true when become_degraded succeeds");
```

---

### Behavior change: `unavail.clear()` removed from write path

The old `OPTIMISTIC` path called `state.backup_dev->unavail.clear(memory_order_release)` on a successful backup write, immediately re-enabling backup writes after a transient error. The new code never clears `unavail` from the write path — only the resync probe tick can clear it now.

Consequence: after a transient backup error, `unavail` remains set until the next probe fires. During that window all dirty-region writes go active-only, widening the resync surface area. For arrays with frequent transient errors this can meaningfully increase resync time.

This is an intentional trade-off to close the stale-read race, but it is undocumented. A short comment in `async_iov` noting why `unavail.clear()` was deliberately removed would help future readers.

---

### No replacement test for new degraded+dirty+available-backup behavior

`write_optimistic.cpp` is correctly deleted. No new test covers the scenario it used to exercise: a degraded array with an available backup where the target region is dirty. Under the new code this must route active-only and call `dirty_region()`; under the old code it was `OPTIMISTIC`. A unit test asserting `backup_write == false` and that the bitmap stays dirty after the write would protect against regressions.

Ilyin-Ilya (changes requested):
## Review (updated after further analysis)

The core approach is correct and the invariant it establishes is elegant: **dirty regions are exclusively owned by the resync task; the I/O path never touches backup for a dirty region.** This makes the two agents' working sets disjoint and eliminates the coordination problem OPTIMISTIC created.

---

## Blocking

### `README.md` line 107 — stale reference to removed feature

The diff updates line 16 but misses line 107, which still reads:
```
- Optimistic write recovery with dirty-bitmap-guided resync
```
Should be removed or replaced (e.g. "Background resync with region-level conflict detection").

---

### `docs/error_codes.md` — OPTIMISTIC row not cleaned up

The error codes table still documents the now-deleted mode:
```
| WRITE | Active and backup both succeed (OPTIMISTIC mode) | ... | Chunk cleaned in bitmap; backup marked available |
```
Must be removed or rewritten for the two-mode (`backup_write=true/false`) model.

---

## Non-blocking

### DEBUG_ASSERT message wording

```cpp
DEBUG_ASSERT(backup_write, "backup_write must not be SKIP when become_degraded succeeds");
```
`backup_write` is a `bool` — "must not be SKIP" is meaningless and will appear verbatim in crash logs. Suggest: `"backup_write must be true when become_degraded succeeds"`.

---

### Missing replacement test for degraded+dirty+available-backup path

`write_optimistic.cpp` tested chunk-aligned dirty writes in degraded mode with backup available. Under the new code that path now takes SKIP (active only, `dirty_region()` called, bitmap stays dirty). A small unit test covering this new behavior would prevent regressions.

---

## Notes on the design (for documentation / commit message)

A few things worth capturing in a comment or the PR description for future readers:

**Why SKIP and not "write to both + stay dirty":**
The natural follow-up question is whether we could write to backup for dirty regions (keeping it more current) while only relying on the resync task to call `__clean`. The answer is no — not as an improvement. Resync reads from active; if an I/O write lands on active during the copy window, resync writes stale data to backup, then Phase 2 correctly skips `__clean` and retries next sweep. The "write to both" I/O work is simply undone by resync. There is no convergence benefit, only extra backup writes. SKIP is the right call.

**Why the 2-phase conflict check is still necessary:**
Even with SKIP the I/O path still updates active for dirty regions. Resync reads from active. Without Phase 2 (`completed_since`), resync could copy a stale snapshot of active to backup and then call `__clean` — leaving backup with old data and the dirty bit cleared, enabling stale reads. Phase 1 (`overlaps`) is an optimization that avoids wasted copies when a write is in-flight; Phase 2 is what enforces correctness.

**`unavail.clear()` deliberately removed:**
OPTIMISTIC used to clear `backup_dev->unavail` on a successful backup write, immediately re-enabling the backup after a transient error. With SKIP, only the resync probe tick can clear `unavail`. This is an intentional trade-off: a transient error now keeps the device skipped until the next probe interval rather than self-healing on the next aligned write. Worth a comment in the code so future readers don't try to add it back.

szmyd:
Thanks @Ilyin-Ilya — addressed in 84f3355 (that push also brings in the `main` merge that resolves the conflict).

**Blocking**
- **`README.md:107`** — removed the stale "Optimistic write recovery" bullet. I deleted rather than replaced it: the *Resync Features* section just below already describes the current region-coordinated resync, so a replacement line would have duplicated it.
- **`docs/error_codes.md`** — removed the OPTIMISTIC row. The degraded outcome it half-described is already covered by the row directly above ("Active succeeds; backup mode SKIP (degraded + unavail/dirty)").

**Non-blocking**
- **DEBUG_ASSERT wording** — now reads `"backup_write must be true when become_degraded succeeds"`. Only the `sync_iov` site still had the old text; the `async_iov` site now asserts on `backup_task.has_value()` instead.
- **Replacement test** — added `SyncIoWriteDegradedDirtySkipsAvailableBackup`. It uses the degraded-via-unclean-shutdown fixture (backup reachable, `unavail` clear, region dirty) and asserts the write routes active-only (`disk_b` data write `Times(0)`) and the region stays dirty. This pins the **`is_dirty`-triggered** skip specifically, which the existing degraded-write tests don't cover — those skip because of `unavail`.

**On `unavail.clear()` (your design note):** we actually went the other way after your review. 5c7f5e4 re-adds `unavail.clear()` on a successful write (both active and backup), so a device that recovers is re-enabled immediately instead of waiting for the idle probe. This is safe w.r.t. the race this PR closes: clearing `unavail` only re-enables backup writes for *clean* regions — a dirty region still forces SKIP via the `is_dirty(addr, len)` check regardless of `unavail`, so the resync task stays the sole authority on cleaning. So the "transient error widens the resync surface until the next probe" concern is resolved by that change rather than by a comment warning future readers off.

Your other notes (why SKIP beats "write-to-both", why the Phase 2 `completed_since` check is the correctness gate) are captured in the PR description and the *Resync Features* section of the README.

src/raid/raid1/raid1.cpp
```diff
@@ -817,12 +810,13 @@ io_result Raid1Disk::sync_iov(uint8_t op, iovec* iovecs, uint32_t nr_vecs, off_t
     }
 
     // WRITE / DISCARD / WRITE_ZEROES: flat replication -- mirrors async_iov write path.
-    auto const is_discard = (op == UBLK_IO_OP_DISCARD || op == UBLK_IO_OP_WRITE_ZEROES);
 
     // Register this write's LBA range in the region tracker so resync skips only the
     // conflicting chunk rather than pausing globally.
     auto _guard = raid1::ResyncWriteGuard{*_resync_task, static_cast< uint64_t >(addr), len};
-    auto const bm = __compute_backup_mode(state, static_cast< uint64_t >(addr), len, is_discard);
+    auto const backup_write =
```
Ilyin-Ilya: claude posted a comment about backup_write duplication in sync/async IOv, potentially we can have a macro for it, but I'm fine if you want to keep as it is. 
Please don't forget to change the version and the changelog to reflect the latest main.

Ilyin-Ilya (dismissed):
the logic looks correct, claude posted a comment about the test, we can address and merge

src/raid/raid1/raid1.cpp
szmyd: Addressed in d026bed:

- **backup_write duplication**: extracted into a typed private helper `__backup_writable(state, addr, len)` (rather than a macro) so the skip-backup-for-dirty-region invariant lives in one documented place and cannot drift between `async_iov` and `sync_iov`. Both call sites now just call it.
- **version + changelog**: already reflects latest main - the branch bumps `0.32.2 -> 0.32.3` (patch, for a fix) with a `[0.32.3]` Fixed entry; I added the `2026-05-25` release date to that heading for consistency with the other entries.

Debug build green, 14/14 test suites pass.

szmyd:
I'll convert the one remaining NiceMock; i missed i thanks!

Ilyin-Ilya (approved):

PR #271, opened 2026-05-24
fix(raid0): guard constructor against capacity underflow, zero max_sectors, and oversized stripe

Ilyin-Ilya:
## Summary

Fixes three construction-time crashes and undefined-behaviour scenarios in `Raid0Disk` (from the static-analysis issue list #269):

- **H2**: throw when device capacity ≤ `stripe_size` — an unsigned subtraction `dev_sectors - stripe_size_sectors` previously underflowed silently, producing a nonsensical volume size.
- **M2**: throw when `max_sectors == 0` after the child-device `min` loop — a zero `max_sectors` made the subsequent `dev_sectors % max_sectors` alignment divide by zero.
- **C1 (validation)**: after loading the superblock (which may override `_stripe_size`), recompute `_stride_width` and validate that `ceil(max_io / stride_width)` ≤ `_max_stripe_cnt=64`. Without this, a loaded stripe_size that is too small for the device's `max_io` would silently overflow the fixed-size `io_array` at runtime.

## Test plan

- [x] `Raid0.DeviceSmallerThanStripeSizeThrows` — H2 guard
- [x] `Raid0.ZeroMaxSectorsThrows` — M2 guard
- [x] `Raid0.StripeToSmallForMaxIoThrows` — C1 validation

### DIFF
--- a/src/raid/raid0/raid0.cpp
+++ b/src/raid/raid0/raid0.cpp
@@ -111,6 +111,8 @@ Raid0Disk::Raid0Disk(boost::uuids::uuid const& uuid, uint32_t const stripe_size_
     // superblock value (the initializer-list computed it from the caller-supplied stripe_size_bytes
     // before any superblock was read, so it may be stale).
     _stride_width = _stripe_size * static_cast< uint32_t >(_stripe_array.size());
+    if (_stripe_size == 0)
+        throw std::runtime_error("Raid0Disk: on-disk superblock delivered zero stripe_size (possible data corruption)");
     // Recompute io_opt_shift to match the corrected stride width.
     our_params.basic.io_opt_shift = ilog2(_stride_width);
 
--- a/src/raid/raid0/raid0.cpp
+++ b/src/raid/raid0/raid0.cpp
@@ -107,11 +107,42 @@ Raid0Disk::Raid0Disk(boost::uuids::uuid const& uuid, uint32_t const stripe_size_
         _stripe_array.emplace_back(std::make_unique< StripeDevice >(std::move(device), sb.value()));
     }
 
+    // Recompute _stride_width in case load_superblock corrected _stripe_size from the on-disk
+    // superblock value (the initializer-list computed it from the caller-supplied stripe_size_bytes
+    // before any superblock was read, so it may be stale).
+    _stride_width = _stripe_size * static_cast< uint32_t >(_stripe_array.size());
+    // Recompute io_opt_shift to match the corrected stride width.
+    our_params.basic.io_opt_shift = ilog2(_stride_width);
+
+    // Validate that the per-stripe iovec count fits in io_array (size _max_stripe_cnt).
+    // Worst case: ceil(max_io / stride_width) iovecs per stripe. With default max_io=512KiB,
+    // _stripe_size=4KiB, N=2, this is exactly 64 = _max_stripe_cnt. Must check AFTER _stride_width
+    // is recomputed above (load_superblock may have corrected _stripe_size).
+    {
+        auto const max_io = static_cast< uint64_t >(our_params.basic.max_sectors) << SECTOR_SHIFT;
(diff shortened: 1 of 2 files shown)


Ilyin-Ilya (commented):
LGTM with two non-blocking concerns.

**H2 / M2 / overflow guards** — all three new checks in `__init_params` are correct. The `dev_sectors <= stripe_size_sectors` guard prevents the silent-zero capacity regression, `max_sectors == 0` avoids the ilog2 UB, and the `(max_io + _stride_width - 1) / _stride_width > _max_stripe_cnt` overflow check is sound.

**`io_opt_shift` stamping** — placing `our_params.basic.io_opt_shift = ilog2(_stride_width)` after the loop is the right spot; `_stride_width` is fully computed by then.

---

**Concern 1 — missing `_stripe_size == 0` guard before `_stride_width` computation**

If on-disk SB corruption delivers `stripe_size = 0`, `_stride_width` becomes 0 and the overflow check immediately divides by zero:

```cpp
if ((max_io + _stride_width - 1) / _stride_width > _max_stripe_cnt) throw ...
```

The H2 guard covers `dev_sectors <= stripe_size_sectors` but not `stripe_size == 0` directly. A `if (_stripe_size == 0) throw ...` before the multiplication would close the gap.

**Concern 2 — `StripeAccum::io_array` stack growth**

`io_array` grew from `std::array<iovec, 16>` to `std::array<iovec, _max_stripe_cnt>` (64). `StripeAccum` lives on the stack inside `__distribute` on the hot I/O path — each instance is now 64 × `sizeof(iovec)` = 1 KiB. Worth watching under stack-depth profiling once wider RAID0 configs land in CI.

Neither concern is a blocker.

src/raid/raid0/raid0.cpp
```diff
@@ -166,15 +199,22 @@ void Raid0Disk::probe_tick(ublksrv_queue const* q) noexcept {
...
+    // Use _max_stripe_cnt as the upper bound: it already caps device count, and with a
+    // minimum stripe_size of logical_block_size the per-stripe iovec count is bounded by
+    // ceil(max_io / (_stripe_size * N)) <= ceil(max_io / _stripe_size) <= _max_stripe_cnt.
     struct StripeAccum {
-        uint64_t io_addr{0};                // Starting address
-        uint32_t nr_vecs{0};                // How many iovecs are valid
-        std::array< iovec, 16 > io_array{}; // The scatter-list
+        uint64_t io_addr{0};                             // Starting address
+        uint32_t nr_vecs{0};                             // How many iovecs are valid
+        std::array< iovec, _max_stripe_cnt > io_array{}; // The scatter-list
```
szmyd: `_max_stripe_cnt` is much bigger than necessary and not the what this should be a function of. This size is needs to accommodate the `_max_tx` wrapping across a given stripe up to "this number" of times. This probably requires a min_stripe size of some reasonable number (maybe 32KiB?); which would then be:
```c++
std::array< iovec, (_max_tx / _min_stripe_size) + 1 > io_array{}; // The scatter-list

File: src/raid/raid0/raid0.cpp [hunk 2]
```diff
@@ -166,15 +199,22 @@ void Raid0Disk::probe_tick(ublksrv_queue const* q) noexcept {
...
     };
     static_assert(_max_stripe_cnt == 64, "dirty_mask must be exactly uint64_t");
 
-    // Then when allocated (once) an array of accumulators for the thread (1-thread per i/o queue)
+    // ~66 KiB per queue thread (_max_stripe_cnt × sizeof(StripeAccum) = 64 × ~1040 B); bounded and intentional.
```
szmyd: Likely `17 x 1040B`ish, not 64

szmyd (changes requested):
Array size should be a function of max_tx and some stipe_size, not the number of stripes.

szmyd (approved):
Great!

PR #272, opened 2026-05-24
fix(raid1/bitmap): overflow guards — max_tx zero-check, uint64_t page_width, relaxed dirty estimate

Ilyin-Ilya:
## Summary

Four correctness fixes in `Bitmap` (from the static-analysis issue list #269):

- **H7 (`init_to` zero `max_pages` guard)**: if a device's `max_tx()` is smaller than `k_page_size`, `max_pages_per_tx()` returns 0 and the write loop would spin forever or skip initialisation silently. Now throws `std::runtime_error` immediately.

- **`sync_to` zero `max_batch` guard**: same class of defect — returns `std::errc::invalid_argument` rather than attempting a zero-length allocation.

- **C3 (`_page_width` widened to `uint64_t`)**: at `chunk_size = 128 KiB` the old `uint32_t` expression `128 KiB × 4 KiB × 8 = 4 GiB` silently wraps to 0, corrupting all range arithmetic with a SIGFPE on the first division. Widening to `uint64_t` fixes the overflow.

- **`clean_region` dirty-estimate update**: replaced a CAS loop with a plain `fetch_sub(min(load, cleared), relaxed)`. The CAS was guarding against TOCTOU underflow from concurrent `clean_region` callers, but `clean_region` is now called from a single thread only. The only concurrent accessor is `dirty_region`, which exclusively does `fetch_add` — so the value at `fetch_sub` time is always ≥ the loaded snapshot, making underflow impossible.

## Test plan

- [x] `Raid1.InitBitmapThrowsWhenDeviceMaxTxTooSmall` — H7: verifies throw when device `max_tx` < `k_page_size`.
- [x] `Raid1.BitmapWith128KiChunkDoesNotOverflow` — C3: verifies construction succeeds without SIGFPE at `chunk_size = 128 KiB`.

### DIFF
--- a/src/raid/raid1/bitmap.cpp
+++ b/src/raid/raid1/bitmap.cpp
@@ -94,6 +94,9 @@ void Bitmap::init_to(std::shared_ptr< ublk_disk > device) {
     RLOGI("Clearing RAID-1 BITMAP [pgs:{}, sz:{}Ki, id:{}] on: {}", _num_pages, _num_pages * k_page_size / Ki, _id,
           *device)
     auto const max_pages = max_pages_per_tx(*device);
+    if (0 == max_pages)
+        throw std::runtime_error(
+            fmt::format("Device max_tx()={} is smaller than bitmap page size ({})", device->max_tx(), k_page_size));
     auto iov = std::unique_ptr< iovec[] >(new iovec[max_pages]);
     if (!iov) throw std::runtime_error("OutOfMemory"); // LCOV_EXCL_LINE
     std::fill_n(iov.get(), max_pages, proto);
@@ -110,6 +113,7 @@ void Bitmap::init_to(std::shared_ptr< ublk_disk > device) {
 io_result Bitmap::sync_to(ublk_disk& device, uint64_t offset) {
     // Allocate iovec array for batching consecutive pages
     auto const max_batch = max_pages_per_tx(device);
+    if (0 == max_batch) return std::unexpected(std::make_error_condition(std::errc::invalid_argument));
     auto iovs = std::unique_ptr< iovec[] >(new iovec[max_batch]);
     if (!iovs) return std::unexpected(std::make_error_condition(std::errc::not_enough_memory)); // LCOV_EXCL_LINE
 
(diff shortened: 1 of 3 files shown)


Ilyin-Ilya (commented):
All three fixes are correct.

**`init_to` zero `max_pages` guard** — throwing on `max_pages == 0` is the right call; the alternative (silently looping forever or leaving the disk uninitialised) is worse.

**`sync_to` zero `max_batch` guard** — returning `invalid_argument` keeps the error visible to the caller. Correct.

**`clean_region` CAS loop** — replaces the `load() + fetch_sub()` TOCTOU that could underflow `_dirty_chunks_est` to `UINT64_MAX`. The loop:

```cpp
auto cur_est = _dirty_chunks_est.load(relaxed);
while (cur_est > 0 &&
       !_dirty_chunks_est.compare_exchange_weak(cur_est, cur_est - std::min(cur_est, cleared),
                                                relaxed, relaxed))
    ;
```

The `std::min(cur_est, cleared)` clamp is the key correctness guard — it prevents the subtraction from going negative even if `cleared > cur_est`. Loop terminates because either CAS succeeds or `cur_est` converges to 0. Sound.

**`_page_width` widened to `uint64_t`** — at `chunk_size = 128 KiB`, the old `uint32_t` expression `128 KiB × 4 KiB × 8 = 4 GiB` silently wraps to 0 and corrupts all range arithmetic. Widening fixes it cleanly.

src/raid/raid1/bitmap.cpp
```diff
@@ -308,10 +311,15 @@ std::tuple< Bitmap::word_t*, uint32_t, uint32_t > Bitmap::clean_region(uint64_t
...
-            std::memory_order_relaxed);
+        auto const cleared = static_cast< uint64_t >(std::popcount(old_word xor (old_word & clear_mask)));
+        // Use a CAS loop to avoid TOCTOU underflow: a plain load() + fetch_sub() pair is not
+        // atomic; a concurrent clean_region() between them can cause _dirty_chunks_est to drop
+        // below `cleared`, making the subtraction wrap to UINT64_MAX.
+        auto cur_est = _dirty_chunks_est.load(std::memory_order_relaxed);
+        while (cur_est > 0 &&
+               !_dirty_chunks_est.compare_exchange_weak(cur_est, cur_est - std::min(cur_est, cleared),
+                                                        std::memory_order_relaxed, std::memory_order_relaxed))
+            ;
```
szmyd: Clean region is single threaded now following the fix to make it the only caller of `clean_region` so we can relax not restrict this function.

szmyd (changes requested):
Relax, not restrict.

szmyd (approved):
:+1:

PR #273, opened 2026-05-24
fix(raid1): superblock bitfield mutex, preserve bitmap age on SB error, sync bitmap when backup is missing

Ilyin-Ilya:
## Summary

Three correctness fixes in `Raid1Disk` (from the static-analysis issue list #269):

- **M5 (data race)**: `clean_unmount`, `read_route`, and `device_b` are bitfields that share one packed byte in `SuperBlock`. Two concurrent `write_superblock` callers (`__become_clean` on the resync thread and `__become_degraded` on an I/O queue thread) each do a compiler-generated non-atomic read-modify-write on that byte, silently corrupting adjacent bits. Fix: add `_sb_fields_lock` (std::mutex) around every site that writes these bitfields, serializing concurrent callers.

- **M6 (data integrity)**: `__become_degraded` was reverting the in-memory `bitmap.age` on superblock write failure. Since the in-memory route CAS had already succeeded, in-memory state was already degraded. Reverting `age` while keeping the new route active made on-disk state appear pre-degradation, causing device-selection corruption on the next restart. Age is now preserved on error; only `unavail` is set so the device is fenced.

- **M11 (regression)**: the destructor skipped `bitmap.sync_to()` when the backup was a missing-leg placeholder (`is_degraded && !backup->is_missing()`). This meant the dirty bitmap was not persisted on clean shutdown, forcing a full resync on the next swap-in. Condition simplified to `is_degraded` — `sync_to()` handles a missing-placeholder backup transparently.

## Test plan

- [x] `Raid1.DegradedMissingBackupSyncsBitmapAtShutdown` — M11 regression: verifies the bitmap page is written to the active device at shutdown even when the backup slot is a missing placeholder.
- M5 and M6 are best covered by ThreadSanitizer (GccThreadSanitize CI job), not unit tests.

### DIFF
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -255,8 +255,12 @@ Raid1Disk::~Raid1Disk() {
     if (!_sb) return;
 
     auto const state = __capture_route_state();
-    // Write out our dirty bitmap
-    if (state.is_degraded && !state.backup_dev->disk->is_missing()) {
+    // Write out our dirty bitmap.
+    // Flush even when the backup is a missing placeholder: the active device always holds the
+    // authoritative bitmap, and persisting it avoids a full resync when the missing disk is
+    // swapped in later. Skipping the flush when backup is missing was the old behaviour and
+    // caused incremental resync to fall back to a full resync unnecessarily (M11).
+    if (state.is_degraded) {
         RLOGI("Synchronizing BITMAP [uuid: {}] to clean device: {}", _str_uuid, *state.active_dev->disk)
         if (auto res = _dirty_bitmap->sync_to(*state.active_dev->disk, sizeof(SuperBlock)); !res) {
             RLOGW("Could not sync Bitmap to device on shutdown, will require full resync next time! [uuid:{}]",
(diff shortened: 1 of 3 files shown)


Ilyin-Ilya (commented):
All three fixes are correct.

**`_sb_fields_lock`** — `clean_unmount`, `read_route`, and `device_b` share one packed byte in the SB struct. Concurrent writes to different bitfields in the same byte are C++ UB even when runtime data races cannot occur (e.g. behind a thread join), because the compiler is free to implement each field write as a read-modify-write on the whole byte. Guarding every `write_superblock` call site with `_sb_fields_lock` eliminates the UB cleanly. The lock is `mutable` so it can be taken in `const` contexts — correct pattern.

**M11 fix: `if (state.is_degraded)` (drop `&& !backup->is_missing()`)** — the old condition skipped the bitmap sync to the active device when the backup was a missing placeholder. That meant a degraded array with a fully-missing disk would not persist its dirty bitmap at shutdown, forcing a full resync on the next replacement instead of an incremental one. The new condition always syncs when degraded — correct invariant.

**Age rollback removal in `__become_degraded`** — reverting `_sb->fields.bitmap.age` on SB write failure was wrong: the in-memory route had already transitioned to degraded, so any subsequent successful SB write would carry the pre-degradation (stale) age, potentially causing `pick_superblock` to prefer the wrong device after a crash. Keeping the incremented age in memory and accepting that a write failure leaves it unsynced is the safer choice.

Ilyin-Ilya (commented):
Thanks for the sharp review. Addressing each point:

**M5 — scope clarification (valid concern)**

You are correct on both sub-points:

1. `auto local = *sb` is an unsynchronised read of the whole struct. `__become_degraded` writes `_sb->fields.bitmap.age` without holding any lock that `write_superblock` also holds (`__become_degraded` explicitly does not take `_ctrl_lock` — see the comment at the top of `__swap_device`). So `__become_clean`'s `write_superblock` reading `*_sb` can race with `__become_degraded`'s `bitmap.age` increment. This race **preexisted** the change — before the local-copy approach, `write_superblock` was writing `read_route`/`device_b` into the shared struct while `__become_degraded` wrote `bitmap.age` to the same struct. Same race, different fields.

2. `_sb->fields.clean_unmount = 0x1` in the destructor is safe: all I/O queue threads and the resync thread are joined before the destructor body executes, so no `write_superblock` call can overlap with it.

What the local-copy fix **does** eliminate is the write-write race between two concurrent `write_superblock` callers (`__become_clean` on the resync thread vs. `__become_degraded` on an I/O queue thread) each doing a non-atomic read-modify-write on the packed bitfield byte shared by `clean_unmount`, `read_route`, and `device_b`. That is the most dangerous form: compilers generate a load-mask-or-store sequence for bitfield writes, so one thread's store silently overwrites the other's bits.

The code comment has been updated to reflect the narrower scope and explicitly calls out the pre-existing `bitmap.age` read-write race. A complete fix for that race would require protecting `bitmap.age` writes (in `__become_degraded` and `__swap_device`) with the same lock held during the struct copy — that is a separate, larger change.

PR #274, opened 2026-05-24
fix(raid1): ban resync_level=0 at construction time

Ilyin-Ilya:
## Summary

- **Root cause**: with `resync_level=0` the expression `(0 * 100 / 32) * 5 = 0` is assigned to
  `copies_left`; the inner while-loop exits immediately on every sweep with no data copied, so the
  resync task spins indefinitely and the array never recovers from degraded state.
- **Fix**: reject `resync_level=0` eagerly in the `MirrorDevice` constructor (same pattern as
  `chunk_size` validation) — throw `std::runtime_error` at startup rather than silently making no
  progress at runtime. Updated the option description from `(0-32)` to `(1-32)`.
- **Conflict resolution**: rebased onto upstream; upstream's two-phase region-tracker rewrite
  (`#253`) already rewrote `__run`, so the old call-site clamp is dropped in favour of the
  construction-time ban.
- **Version**: bumped to `0.32.2` (behaviour change: previously-silent misconfiguration now throws
  at startup).

## Test plan

- `TEST(Raid1, ZeroResyncLevelThrows)` in `misc/edge_cases.cpp` verifies construction throws with
  `resync_level=0`.
- Dedicated CTest entry `Raid1ZeroResyncLevelThrows` runs `test_raid1 --resync_level=0
  --gtest_filter=Raid1.ZeroResyncLevelThrows` so the global SISL option is set before any
  `MirrorDevice` is constructed; `--gtest_filter` isolates the run from other tests that expect a
  valid level.
- LCOV exclusion annotations removed from the now-covered validation path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.32.2] - 2026-05-24
+
+### Changed
+- **Ban `resync_level=0`**: construction now throws `std::runtime_error` if `resync_level` is set
+  to 0. A zero level silently produced zero `copies_left`, causing the resync loop to stall with
+  no forward progress. Valid range is 1–32.
+
 ## [0.32.1] - 2026-05-24
 - Some fixes for gcc-16 compilation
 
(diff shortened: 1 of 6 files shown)


Ilyin-Ilya (commented):
Fix is correct.

`std::max(1U, ...)` prevents `copies_left` from reaching 0 when `resync_level == 0`, which previously caused the resync task to stall with nothing to copy. The floor of 1 ensures at least one chunk is always attempted per iteration, so progress is guaranteed regardless of level.

One nit: no test was added for the `resync_level = 0` code path. A test that sets `resync_level = 0` and verifies the resync task makes forward progress (rather than spinning) would make the fix explicit and prevent regression.

szmyd (approved):
:+1:

PR #275, opened 2026-05-24
fix(target,driver): recovery return-on-failure, catch-all exception handler, sem_destroy, RAII fd guard, ilog2 zero-guards

Ilyin-Ilya:
## Summary

Five defensive fixes across `ublkpp_tgt.cpp` and `fs_disk.cpp` (from the static-analysis issue list #269):

**`src/target/ublkpp_tgt.cpp`**

- **H5**: `ublksrv_ctrl_start_recovery` failure was logged but execution continued, reaching subsequent recovery steps in an undefined state. Now returns `std::errc::operation_not_permitted` immediately on failure.
- **H6**: the io_uring completion handler only caught `std::exception`; any other type (`std::bad_alloc`, a third-party exception, etc.) would propagate through the `noexcept` coroutine boundary and terminate the process. Added a `catch (...)` fallback that completes the IO with `-EIO`.
- **M8**: `sem_destroy` was never called on the queue semaphore created in device startup, leaking OS resources on every device teardown.

**`src/driver/fs_disk.cpp`**

- **M7**: `FSDisk` opened `_fd` and then could `throw` during `fstat`, `ioctl`, or `fcntl` without closing it, leaking the file descriptor. Added an RAII `FdGuard` that closes `_fd` on any exception path; `release()` is called only after the constructor completes successfully.
- **L3**: `ilog2(0)` is undefined behaviour (typically a SIGFPE). Added explicit zero-checks for `lbs` and `pbs` before passing them to `ilog2` on both the block-device and regular-file code paths.

## Test plan

These fixes cover OS-level failure modes (fd leak on exception, semaphore teardown, ioctl error handling) and ublksrv integration paths that are not practical to exercise in unit tests. Covered by code review and AddressSanitizer/ThreadSanitizer CI runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.32.5] - 2026-05-26
+
+### Fixed
+
+- **H5 (target)**: `ublksrv_ctrl_start_recovery` failure was logged but execution continued into subsequent recovery steps in an undefined state. Now returns `std::errc::operation_not_permitted` immediately on failure.
+- **H6 (target)**: the io_uring completion handler only caught `std::exception`; any other thrown type would propagate through the `noexcept` coroutine boundary and terminate the process. Added a `catch (...)` fallback that completes the IO with `-EIO`.
+- **M8 (target)**: `sem_destroy` was never called on the queue semaphore, leaking an OS resource on every device teardown.
+- **M7 (driver)**: `FSDisk` could `throw` during `fstat`/`ioctl`/`fcntl` without closing `_fd`, leaking the file descriptor. Added an RAII `FdGuard` that closes `_fd` on any exception path.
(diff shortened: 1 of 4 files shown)


Ilyin-Ilya (commented):
All five fixes are correct and well-structured.

**H5 — RAII `FdGuard`** — the pattern is correct: `FdGuard` closes `_fd` on destruction unless `release()` is called, and `release()` is called only at the end of a fully successful constructor. Exception safety is clean.

**H6 — `lbs == 0 || pbs == 0` guards before `ilog2`** — `ilog2(0)` is UB on most implementations (GCC `__lg` returns -1, which then shifts by a negative amount). The explicit zero-check throwing `std::system_error(EINVAL)` is the right response for a block device reporting a zero block size.

**M8 — `ublksrv_ctrl_start_recovery` return value checked** — the previous code assigned the return to a local `ret` inside a dangling `else` branch and discarded it. Splitting into a checked call with a separate error path is correct.

**M7 — `catch (...)` handler** — catching and logging the unknown exception prevents silent termination propagation from the recovery path. Good.

**L3 — `sem_destroy`** — POSIX semaphores must be destroyed after use; omitting `sem_destroy` is a resource leak on some implementations. Correct fix.

szmyd (dismissed):
:+1:

szmyd (commented):
:+1:

szmyd (approved):
Looks good

PR #278, opened 2026-05-25
fix(raid1): snapshot iovecs in __failover_read_async to prevent stale-iovec corruption

szmyd:
## Summary

- Snapshots \`iovecs\` into a coroutine-frame-local \`std::array<iovec, 16>\` at the top of \`__failover_read_async\`, before any \`co_await\`, to prevent use of stale RAID0 \`thread_local sub_cmds\` data in the failover read path
- Adds \`MockUblksrv::iov_ref()\` accessor to expose the caller-side iovec for test mutation
- Adds regression test \`FailoverUsesSnapshotIovecNotStaleCopy\` that simulates the race and verifies the failover uses the original \`iov_base\` and \`iov_len\`

## Root Cause

In RAID10, \`RAID0::__distribute\` populates \`thread_local sub_cmds\` and passes \`iovecs\` pointing into it when calling \`RAID1::async_iov\`. Once \`__failover_read_async\` suspends at \`co_await primary_task\`, concurrent I/Os on the same queue thread can overwrite \`sub_cmds\`, leaving the \`iovecs\` pointer stale by the time the failover SQE is submitted. The failover then reads the wrong number of bytes into the wrong buffer.

Confirmed via raw-block fio logs showing \`r > len\` on failover reads (e.g. \`len:4096\` but \`r=131072\`) immediately after a SIGKILL'd AM caused multiple simultaneous in-flight read failures.

The existing \`io_uring_submit\` guard in \`RAID0::async_iov\` only protects the initial SQEs - the failover SQE is submitted much later, after the primary fails and \`sub_cmds\` has been overwritten.

## Why SIGTERM doesn't reproduce

With SIGTERM (30s grace), AMs flush and close iSCSI connections gracefully. No reads are in-flight when the failure is detected, so the failover path is never exercised while \`sub_cmds\` is in a stale window.

## Implementation notes

- \`std::array<iovec, 16>\` is used (not \`std::vector\`) to avoid an extra heap allocation; the coroutine frame is already heap-allocated. The bound of 16 matches \`StripeAccum::io_array\` in RAID0.
- \`DEBUG_ASSERT_LE(nr_vecs, 16u)\` guards the copy in debug builds.
- The regression test runs in a \`std::thread\` to isolate \`thread_local last_read\` state from other tests on the main gtest thread (same pattern used by other \`AsyncRaid1Fixture\` tests).

## Test plan

- [x] CI passes (GccAddressSanitize, GccThreadSanitize, GccCoverage, ClangRelease)
- [x] \`ctest -R Raid1.*FailoverStale\` passes
- [x] \`am_resync\` resiliency test passes with SIGKILL pod restarts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.32.4] - 2026-05-25
+### Fixed
+- raid1: Make stable copy of iovecs in __failover_read
+
 ## [0.32.3] - 2026-05-25
 ### Fixed
 - raid1: Remove Optimistic write path which can now race with region tracked resync.
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=2.0"
 
 class UBlkPPConan(ConanFile):
     name = "ublkpp"
-    version = "0.32.3"
+    version = "0.32.4"
 
     homepage = "https://github.com/szmyd/ublkpp"
     description = "A UBlk library for CPP application"
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -661,7 +661,11 @@ disk_task< int > Raid1Disk::__failover_read_async(ublksrv_queue const* q, ublk_i
     // Snapshot before first co_await: in RAID10, iovecs points into RAID0's thread_local
     // sub_cmds. Once we suspend, concurrent I/Os on the same queue thread can overwrite
     // sub_cmds, making iovecs stale when the failover SQE is submitted.
-    auto iov_snap = std::vector< iovec >(iovecs, iovecs + nr_vecs);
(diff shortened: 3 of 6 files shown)


Ilyin-Ilya:
**Addendum — the test is checking the wrong field**

After re-reading: the dangerous half of the stale-iovec corruption isn't `iov_len` — it's `iov_base`.

When RAID0 starts I/O B and overwrites `sub_cmds`, `iovecs[0].iov_base` no longer points to I/O A's output buffer. It points to **I/O B's output buffer**. The failover SQE then tells the kernel to read disk data into I/O B's buffer. Both I/Os corrupt each other:

- I/O A's user buffer is never filled → I/O A returns stale or zeroed data
- I/O B's user buffer gets overwritten with I/O A's disk content → I/O B returns wrong data

The regression test captures only `iov->iov_len` in the `WillOnce` callback and asserts it equals the original 4 KiB. `iov->iov_base` is not checked at all. A test that verifies `iov_base` still points to the original request buffer (and not to I/O B's buffer) would actually pin the dangerous invariant. As written, the test guards against the `r > len` symptom observed in fio logs, not against the silent wrong-data path.

The snapshot fix (`iov_snap = std::vector<iovec>(iovecs, iovecs + nr_vecs)`) does protect both fields correctly — `iov_base` and `iov_len` are both captured at call time. The fix is right; the test is incomplete.

Ilyin-Ilya (approved):

PR #280, opened 2026-05-26
fix(tests): thread-safety, dead code, vacuous assertions, queue routing

Ilyin-Ilya:
### H8 investigation — equal-ages asymmetric clean_unmount: no behavioral change needed

Investigated the scenario where both superblocks have the same age but one has `clean_unmount=0`. The conclusion is that this requires no special handling in `pick_superblock`:

Equal ages guarantee both devices have identical data. Any data-write failure immediately degrades the array and causes the ages to diverge — the surviving device's superblock advances independently. The only way to reach equal ages + asymmetric `clean_unmount` is a failed shutdown write of `clean_unmount=1`. No data write can produce it, so there is nothing to resync.

`clean_unmount=0` is only meaningful in `__init_bitmap_and_degraded_route` when the route is already non-EITHER (the `route != EITHER && clean_unmount==0` branch that triggers a full bitmap dirty on a crash-during-resync). That case is already handled correctly by the on-disk route value written by `__become_active`.

Cases 3 and 5 in `pick_super.cpp` are updated to assert `EITHER` (the correct on-disk value for a previously healthy array) with comments explaining the invariant. A code comment is added to `pick_superblock` to prevent future regressions of adding a `read_route` override here.

---

### Additional fixes (from issue #269 audit)

- **M4 (bitmap.cpp)**: Move `loaded_from_disk.store(false, release)` to before the bit-write loop. Old ordering allowed a concurrent `sync_to` to observe `loaded_from_disk=true` after new bits were already in memory, skip the page, and lose those bits on crash. New ordering is self-correcting: if `sync_to` captures an older snapshot, `loaded_from_disk` is already false so the next sync cycle flushes the new bits.
- **`raid1.cpp` + `raid1_resync_task.cpp`**: `unavail.test_and_set(memory_order_acquire)` → `memory_order_acq_rel` (5 sites). `acquire`-only on an RMW provides no release semantics on the store half, so threads observing `unavail=true` would not necessarily see preceding writes. `acq_rel` fixes this.
- **M9 (fs_disk.cpp)**: `UINT32_MAX == discard_granularity` guard was dead code — the struct is zero-initialized. Replaced with a zero-check so DISCARD is correctly un-advertised when granularity is unset.
- **M12 (failover_retry_during_swap.cpp)**: Move all `EXPECT_CALL` registrations to the main thread before spawning. GMock's `EXPECT_CALL` is not thread-safe when called concurrently with active mock invocations — TSAN flags this.
- **M13 (mock_ublksrv.cpp)**: Route IOs to `_queues[tag % nr_queues]` instead of always `_queues[0]` so per-queue behaviour is exercised in multi-queue tests.
- **M14 (replica_states.cpp)**: Call `wait_for_clean_state` after `toggle_resync(true)` — the helper was defined but had zero call sites.
- **M15 (pick_super.cpp)**: Replace zero-initialized `read_route` fields with explicit enum casts so assertions are non-vacuous. Add Case 6 (initialises `deva_sb` with `DEVB`, asserts result is `DEVA`) to prove `pick_superblock` actively overwrites the field rather than preserving it.
- **M16 (io_during_swap.cpp)**: Track `device_b` writes after the swap via an atomic counter and assert zero. `AnyNumber()` previously hid post-swap routing bugs.
- **L5 (test_disk.hpp)**: Remove dead `make_io_data`/`remove_io_data` — zero call sites, `new` without `delete`, non-atomic `static int` tag counter.

## Test plan

- [ ] CI passes all four jobs (GccAddressSanitize, GccThreadSanitize, GccCoverage, ClangRelease)
- [ ] TSAN job no longer flags the `FailoverRetryDuringSwap` EXPECT_CALL race
- [ ] `ToggleResyncDisable` passes with the new `wait_for_clean_state` assertion
- [ ] `WriteDuringSwap` catches post-swap routing errors via the new counter assertion
- [ ] `PickSuper` Case 6 fails if `pick_superblock` does not actively write `DEVA`


src/raid/raid1/raid1_resync_task.cpp
```diff
@@ -377,7 +377,7 @@ bool Raid1ResyncTask::probe_mirror(MirrorDevice& mirror, uint64_t reserved_size)
         mirror.unavail.clear(std::memory_order_release);
         return true;
     }
-    mirror.unavail.test_and_set(std::memory_order_acquire);
+    mirror.unavail.test_and_set(std::memory_order_acq_rel);
```
Ilyin-Ilya, line 380: this type of changes is safe because on x86 they will compile to the same instruction, so no diff between memory_order_acquire and memory_order_acq_rel.

Ilyin-Ilya, line 380: probably it's also a good change because we spawn this mirror.unavail.test_and_set everywhere across the code, it's hard to maintain sync relation.

src/raid/raid1/raid1_superblock.cpp
```diff
@@ -26,8 +26,18 @@ raid1::SuperBlock* pick_superblock(raid1::SuperBlock* dev_a, raid1::SuperBlock*
...
         return dev_a;
-    } else if (dev_a->fields.clean_unmount != dev_b->fields.clean_unmount)
-        return dev_a->fields.clean_unmount ? dev_a : dev_b;
+    } else if (dev_a->fields.clean_unmount != dev_b->fields.clean_unmount) {
+        // Ages are equal but one device had an unclean shutdown. Route reads to the clean device
+        // so __init_bitmap_and_degraded_route enters the degraded path and loads the dirty bitmap
+        // from the authoritative replica. Without setting read_route here the array opens as
+        // healthy (EITHER) and the unclean device's in-flight writes are silently ignored.
+        if (dev_a->fields.clean_unmount) {
+            dev_a->fields.read_route = static_cast< uint8_t >(read_route::DEVA);
```
Ilyin-Ilya: basically here we do the same as for other conditional branches, but for some reason it's just forgotten in that else if.

szmyd (approved):
lgtm

PR #281, opened 2026-05-26
fix(raid0): SB version guard, stride_width widening, power-of-2 check, submit-failure abort

Ilyin-Ilya:
## Summary

Fixes four bugs in the RAID0 implementation. Closes #269.

### H1: Deadlock + UAF on \`io_uring_submit\` failure

When \`io_uring_submit\` fails in \`async_iov\`, child coroutines are suspended at \`co_await *cqe_state\` waiting for CQEs that will never arrive. The original drain loop (\`for t in stripe_tasks: co_await t\`) deadlocks because \`hot_task::await_ready()\` returns false and the child never resumes.

Additionally, the staged SQEs remain in the ring. The next \`io_uring_submit\` call from any subsequent IO on the same queue thread flushes them, delivering CQEs that reference recycled \`cqe_state\` pointers — a use-after-free.

**Fix:** Retry \`io_uring_submit\` once (the ring may be transiently full). If the retry succeeds, fall through to the normal \`co_await\` drain loop — the child coroutines complete normally when their CQEs arrive. If the retry also fails, the ring is genuinely broken (EBADF, ENOMEM, …); \`RELEASE_ASSERT_GE\` aborts the process. Aborting is simpler than synthetic completion and eliminates the stale-SQE UAF: any SQEs still staged in the ring are never flushed because the process no longer runs.

### M3: Future superblock versions not rejected

\`load_superblock\` checked \`SB_VERSION > sb_ver\` (migrate old → new) but had no upper-bound check. An on-disk superblock written by a newer version of the code would be silently processed with the wrong field layout, potentially causing silent data corruption.

**Fix:** Added \`if (sb_ver > k_sb_version)\` guard that returns \`std::errc::not_supported\`, causing the constructor to throw \`std::runtime_error\`. \`k_sb_version\` moved from a local \`constexpr\` in \`raid0.cpp\` to \`raid0_impl.hpp\` so tests can reference it directly (\`k_sb_version + 1\`). New unit test \`FutureSuperblockVersionThrows\` covers this case.

### L1: \`_stride_width\` uint32_t overflow

\`_stride_width\` was \`uint32_t\`. For large configs (e.g. 128 MiB stripe × 64 disks = 8 GiB), the product silently overflowed, corrupting IO routing calculations in \`__distribute\`, \`sync_iov\`, and \`async_iov\`.

**Fix:** Widened \`_stride_width\` to \`uint64_t\` throughout: member declaration, initializer-list, post-SB-load recomputation, \`DEBUG_ASSERT_GT\` sentinel. Also widened \`stride_width\` parameters in \`next_subcmd\`/\`merged_subcmds\` in \`raid0_impl.hpp\`; existing implicit \`uint32_t → uint64_t\` promotions at call sites are safe (widening).

### L2: Non-power-of-2 \`stripe_size_bytes\` not rejected

\`ilog2\` rounds down for non-power-of-2 inputs (e.g. 6 KiB → 12, treated as 4 KiB). The wrong \`physical_bs_shift\` and \`io_opt_shift\` were computed silently, producing incorrect geometry.

**Fix:** Added construction-time check \`stripe_size_bytes & (stripe_size_bytes - 1)\` that throws \`std::invalid_argument\`. The check runs before the device loop so no mock calls occur. New unit test \`NonPowerOfTwoStripeSizeThrows\` covers this case.

## Test plan

- [ ] \`NonPowerOfTwoStripeSizeThrows\` — L2: constructor rejects 6 KiB stripe size before any device I/O
- [ ] \`FutureSuperblockVersionThrows\` — M3: constructor rejects SB version \`k_sb_version + 1\` (future format)
- [ ] All existing \`Raid0*\` unit tests continue to pass (no regressions)
- [ ] CI: GccAddressSanitize, GccThreadSanitize, GccCoverage, ClangRelease

### DIFF
--- a/src/raid/raid0/raid0.cpp
+++ b/src/raid/raid0/raid0.cpp
@@ -5,6 +5,7 @@
 #include <ublksrv.h>
 #include <ublksrv_utils.h>
 
+#include <ublkpp/lib/cqe_state.hpp>
 #include <ublkpp/lib/disk_task.hpp>
 #include <ublkpp/lib/ublk_disk.hpp>
 
@@ -41,7 +42,7 @@ class Raid0Disk : public ublk_disk {
     std::vector< std::unique_ptr< StripeDevice > > _stripe_array;
 
     uint32_t _stripe_size{0};
-    uint32_t _stride_width{0};
+    uint64_t _stride_width{0};          // L1: widened to uint64_t; large configs (e.g. 128MiB × 64) overflow uint32_t
     uint32_t _max_iovecs_per_stripe{0}; // ceil(max_tx / stripe_size): max iovecs a single stripe accumulates
 
     io_result __distribute(iovec* iov, uint64_t addr, auto&& func) const;
@@ -70,6 +71,11 @@ Raid0Disk::Raid0Disk(boost::uuids::uuid const& uuid, uint32_t const stripe_size_
         ublk_disk(), _stripe_size(stripe_size_bytes), _stride_width(_stripe_size * disks.size()) {
     if (disks.empty()) throw std::invalid_argument("Raid0Disk: at least one disk required");
     if (stripe_size_bytes == 0) throw std::invalid_argument("Raid0Disk: stripe_size_bytes must be non-zero");
+    // L2: ilog2 rounds down for non-power-of-2 inputs, producing wrong geometry silently.
+    // E.g. stripe_size=6KiB would use ilog2→12, treating it as 4KiB. Reject early.
+    if (stripe_size_bytes & (stripe_size_bytes - 1))
+        throw std::invalid_argument(
+            fmt::format("Raid0Disk: stripe_size_bytes ({}) must be a power of 2", stripe_size_bytes));
     if (disks.size() > _max_stripe_cnt)
         throw std::invalid_argument(
             fmt::format("Raid0Disk: too many disks ({}), max is {}", disks.size(), _max_stripe_cnt));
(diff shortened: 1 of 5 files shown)


src/raid/raid0/raid0.cpp
```diff
@@ -313,12 +319,19 @@ disk_task< int > Raid0Disk::async_iov(ublksrv_queue const* q, ublk_io_data const
...
+    // probability path. Aborting also prevents a stale-SQE UAF: the staged SQEs left in the ring
+    // would be flushed by the next io_uring_submit on this queue, delivering CQEs into an
+    // already-freed coroutine frame.
     if (q && q->ring_ptr && io_uring_submit(q->ring_ptr) < 0) [[unlikely]] { // LCOV_EXCL_START
-        for (auto& t : stripe_tasks)
-            co_await t;
-        co_return op == UBLK_IO_OP_READ ? -EAGAIN : -EIO;
+        if (io_uring_submit(q->ring_ptr) < 0) {
+            RLOGE("io_uring_submit failed twice; ring is broken ({}). Aborting.", strerror(errno))
+            std::abort();
```
szmyd:
```
RELEASE_ASSERT_GE(io_uring_submit(q->ring_ptr), 0, "io_uring_submit failed twice; ring is broken ({}). Aborting.", strerror(errno))
```

szmyd (commented):
The retry + synthetic completion correctly prevents the deadlock, and the _waiter null-check in run_queue_loop:127 (if (auto h = std::exchange(state->_waiter, {})) h.resume()) confirms no double-resume. One remaining concern:

The stale SQEs are still in the ring after both submits fail. 
The next io_uring_submit_and_wait_timeout in run_queue_loop will flush them. Their CQEs could land in the same io_uring_for_each_cqe batch as a FETCH command for a new I/O on the same tag. If the FETCH CQE is processed first, __handle_io_async:277 calls _pool.clear(), destroying the old cqe_state objects. When the stale device CQE is then processed, state->_result_ready = true writes into freed memory - the _waiter null-check is never reached.

  The probability is low (persistent io_uring_submit failure is already very rare, and the CQE ordering within the batch adds another condition), so this may be acceptable as-is. A complete fix would need to cancel the staged SQEs via io_uring_prep_cancel before returning. Is that something worth pursuing here, or is the probability low enough to live with?

szmyd (approved):
lgtm

PR #282, opened 2026-05-26
fix(raid1): bitfield race, age revert, bitmap flush, device size guard

Ilyin-Ilya:
## Summary

Four RAID1 state-management bugs fixed, each with a unit test:

- **M5 (bitfield race):** `clean_unmount`, `read_route`, and `device_b` share a single `uint8_t` in `SuperBlock::fields`. The destructor's bare `_sb->fields.clean_unmount = 0x1` is an unprotected read-modify-write that races with `write_superblock()` calls from I/O paths. Fixed by wrapping the write in `_ctrl_lock` (the same mutex that serializes `__swap_device`), which is uncontended at destruction time since ublksrv queues are torn down before the disk is released.

- **M6 (age revert on SB write failure):** When `__become_degraded` bumps `bitmap.age` and then fails to persist it, the old code reverted the age in memory. The reverted (pre-degradation) age is then written by the next successful SB write (e.g. at shutdown), making the degraded device appear as new as the active one on next restart — causing wrong device selection. Fixed by removing the rollback: the bumped age must stay in memory.

- **M11 (bitmap not flushed with missing backup):** The destructor's `sync_to` call was skipped when `backup_dev->disk->is_missing()`, so a clean shutdown after a hot-swap-to-placeholder left no dirty bitmap on the active device. Next startup forced a full resync instead of the correct incremental one. Fixed by removing the `is_missing()` guard — the active device always gets its bitmap flushed when the array is degraded.

- **L4 (device too small):** After computing `_reserved_size` in `__init_params`, there was no check that the raw device is large enough to hold it. A too-small device silently underflowed `dev_sectors` and then triggered a misleading "exceeds SuperBitmap max capacity" error from the `Bitmap` constructor. Added an explicit size check with a clear error message ("device too small: X bytes < reserved Y bytes").

## Test plan

- [x] `Raid1.DeviceTooSmallThrows` — L4: verifies that constructing with a 64 MiB device (< ~126 MiB reserved region) throws `std::runtime_error` containing "device too small"
- [x] `Raid1.BecomeDegradedAgeNotRevertedOnWriteFailure` — M6: verifies that after `__become_degraded`'s SB write fails, the age written at clean shutdown equals 1 (the degradation bump), not 0 (the reverted value)
- [x] `Raid1.BitmapFlushedToActiveWhenBackupMissing` — M11: verifies that bitmap pages are written to the active device at shutdown even when the backup is a missing placeholder
- [x] All existing tests should continue to pass: `MissingRemountA/B`, `UncleanShutdownWhileDegraded`, etc. — the M11 fix calls `sync_to` on a clean bitmap (no dirty pages → zero extra writes), so write counts are unchanged

### DIFF
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -126,8 +126,7 @@ void Raid1Disk::__init_params() {
     our_params.basic.dev_sectors = k_max_user_data >> SECTOR_SHIFT;
 
     // Now find the what size we should actually set based on the smallest provided device
-    for (auto device_array = std::set< std::shared_ptr< ublk_disk > >{_device_a->disk, _device_b->disk};
-         auto const& device : device_array) {
+    for (auto const& device : std::array{_device_a->disk, _device_b->disk}) {
         if (!device->direct_io()) {
             RLOGW("Device {} does not support O_DIRECT - RAID-1 will use buffered I/O (backend caching not bypassed!)",
                   device)
@@ -164,8 +163,7 @@ void Raid1Disk::__init_params() {
     // Without this check a too-small device silently produces an integer underflow in the
     // dev_sectors adjustment below and a confusing "exceeds SuperBitmap max capacity" error
     // from Bitmap's constructor later.
-    for (auto device_array = std::set< std::shared_ptr< ublk_disk > >{_device_a->disk, _device_b->disk};
-         auto const& device : device_array) {
+    for (auto const& device : std::array{_device_a->disk, _device_b->disk}) {
         if (device->is_missing()) continue; // placeholder — no physical capacity to check
         auto const dev_bytes = device->capacity();
         if (dev_bytes < _reserved_size) {
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -160,6 +160,22 @@ void Raid1Disk::__init_params() {
                                          : (static_cast< uint64_t >(our_params.basic.max_sectors) << SECTOR_SHIFT);
     _reserved_size += ((our_params.basic.dev_sectors << SECTOR_SHIFT) - _reserved_size) % align;
 
+    // Bug L4: Validate that every member device is large enough to hold the reserved region.
+    // Without this check a too-small device silently produces an integer underflow in the
+    // dev_sectors adjustment below and a confusing "exceeds SuperBitmap max capacity" error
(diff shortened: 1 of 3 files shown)


src/raid/raid1/raid1.cpp
```diff
@@ -284,7 +304,16 @@ Raid1Disk::~Raid1Disk() {
         }
         RLOGI("Synchronized: [uuid: {}]", _str_uuid)
     }
-    _sb->fields.clean_unmount = 0x1;
+    // Bug M5: All three bitfields (clean_unmount, read_route, device_b) share a single byte.
+    // Writes to any of them are read-modify-write operations that race with write_superblock()
+    // calls from I/O paths.  Guard the in-memory update with _ctrl_lock — the same lock that
+    // serializes __swap_device — to eliminate the undefined-behaviour window.  By the time the
+    // destructor runs the ublksrv queue threads are torn down (no new I/Os), so this lock is
+    // uncontended; its purpose is to act as a compiler/CPU barrier and document the invariant.
```
Ilyin-Ilya, line 310: The destructor is only reached after the ublksrv queue threads are torn down, so no I/O is in flight at this point and there is no concurrent `write_superblock()` call to race with. The `scoped_lock` is unnecessary — drop it. The code comment explaining the invariant is correct and should stay.

src/raid/raid1/raid1.cpp
```diff
@@ -160,6 +160,22 @@ void Raid1Disk::__init_params() {
                                          : (static_cast< uint64_t >(our_params.basic.max_sectors) << SECTOR_SHIFT);
     _reserved_size += ((our_params.basic.dev_sectors << SECTOR_SHIFT) - _reserved_size) % align;
 
+    // Bug L4: Validate that every member device is large enough to hold the reserved region.
+    // Without this check a too-small device silently produces an integer underflow in the
+    // dev_sectors adjustment below and a confusing "exceeds SuperBitmap max capacity" error
+    // from Bitmap's constructor later.
+    for (auto device_array = std::set< std::shared_ptr< ublk_disk > >{_device_a->disk, _device_b->disk};
```
Ilyin-Ilya: Using `std::set` here is overkill — it heap-allocates a balanced BST just to iterate two known pointers. Replace with `std::array`:\n```cpp\nfor (auto const& device : std::array{_device_a->disk, _device_b->disk}) {\n```\nSame iteration, zero overhead.

PR #283, opened 2026-05-26
fix(raid1): lock-free superblock writes, bitmap flush, device size guard

Ilyin-Ilya:
Consolidates and supersedes PRs #273, #277, #282. Fixes three RAID1 state-management bugs from issue #269 (M5, M11, L4).

---

## M5 — Superblock TSan race: non-atomic read of `superbitmap_reserved`

**The race.** The old `write_superblock` passed `_sb` directly as `iov_base`:

```cpp
// old code
sb->fields.read_route = static_cast<uint8_t>(read_route);
if (device_b) sb->fields.device_b = 1;
auto iov = iovec{.iov_base = sb, .iov_len = sb_size};  // sb == _sb.get()
device.sync_iov(UBLK_IO_OP_WRITE, &iov, 1, 0UL);
sb->fields.device_b = 0;  // reset after write
```

This makes the kernel read all 4 KiB of `_sb` — including `superbitmap_reserved` at offset 74 — as a plain non-atomic memory load. Concurrently, IO coroutines call `SuperBitmap::set_bit` → `std::atomic_ref<uint8_t>::fetch_or` on those exact bytes. A non-atomic load racing with an ongoing atomic RMW is undefined behaviour; this is what TSan detects.

**Why concurrent `write_superblock` callers are not the issue.** The state machine gates each path via CAS: `__become_degraded` opens with `CAS(EITHER → DEVA/DEVB)` and only reaches `write_superblock` if it wins from `EITHER`. `__become_clean` only runs when `route != EITHER`. The same CAS argument applies to `__swap_device`. Two `write_superblock` callers cannot overlap — the competing paths are structurally mutually exclusive. The race is with IO coroutines writing to `superbitmap_reserved`, not with other state-machine callers.

**Fix (lock-free).** `write_superblock` now takes `const SuperBlock*` and never mutates `_sb`. It builds a stack-local `alignas(4096) SuperBlock local{}` buffer and copies only the first 74 bytes of `_sb` (header + fields, stopping before `superbitmap_reserved`). `local.fields.read_route` and `local.fields.device_b` are set on the local copy only; the shared `_sb` is never passed as `iov_base`.

The partial `memcpy` eliminates the race: the local buffer is not shared with any other thread, so no concurrent `atomic_ref` operation can overlap the iov read.

**Destructor path.** The destructor passes `include_superbitmap=true` to `write_superblock`, which copies all 4096 bytes including the live superbitmap. This is safe because `_resync_task->stop()` has already halted all IO coroutines before the destructor reaches this call — no concurrent `set_bit` is possible. This also enables the next startup to do a targeted scan (loading only dirty bitmap pages) rather than a full scan.

---

## M11 — Bitmap not flushed when backup is a missing placeholder

**The scenario.** After a hot-swap where the removed disk is replaced with `make_missing_disk()`, the array is degraded with a missing placeholder as backup. User writes dirty bitmap chunks on the active device. On clean shutdown the old destructor checked:

```cpp
if (state.is_degraded && !state.backup_dev->disk->is_missing())
```

The `is_missing()` guard caused `sync_to()` to be **skipped** entirely. Next startup with the newly swapped-in disk saw no dirty bitmap → fell back to a full resync instead of an incremental one.

**Fix.** Remove the `!is_missing()` guard. Condition is now simply `if (state.is_degraded)`. `sync_to()` is always safe: if there are no dirty pages it writes nothing extra.

**Test.** `BitmapFlushedToActiveWhenBackupMissing`: opens a degraded array with a missing backup, issues a write to dirty one bitmap chunk, then verifies at shutdown that a write to the bitmap region (offset ∈ [4096, reserved_size)) was issued to the active device.

---

## L4 — Device too small to hold reserved region

**The problem.** `__init_params()` computed `_reserved_size` then immediately subtracted it from `dev_sectors` without checking the device was large enough. A device smaller than `_reserved_size` caused an unsigned underflow in `dev_sectors`, producing an enormous phantom capacity. The error surface was a confusing "exceeds SuperBitmap max capacity" exception thrown much later in `Bitmap`'s constructor.

**Fix.** After `_reserved_size` is fully computed and aligned, check against `dev_sectors` — which already holds the minimum of both physical devices from the earlier discovery loop. Checking the minimum once is sufficient: the larger device is always ≥ the minimum and passes trivially.

```cpp
auto const min_dev_bytes = our_params.basic.dev_sectors << SECTOR_SHIFT;
if (min_dev_bytes < _reserved_size)
    throw std::runtime_error(
        fmt::format("device too small: {} bytes < reserved {} bytes", min_dev_bytes, _reserved_size));
```

**Test.** `DeviceTooSmallThrows`: constructs with 64 MiB devices (well below the minimum reserved region for any supported configuration) and asserts a `std::runtime_error` containing "device too small".

---

## Test plan

- [x] `Raid1.DeviceTooSmallThrows` — L4
- [x] `Raid1.BitmapFlushedToActiveWhenBackupMissing` — M11 (clean shutdown with missing backup flushes bitmap)
- [x] `Raid1.MissingDevicePreviouslyDegradedUncleanDirtiesAll` — M11 (unclean degraded path dirties all)
- [x] `Raid1.SuperbitmapNonemptyReturnsFalseOnCleanBitmap` — M5 unit test for `superbitmap_nonempty()` helper
- [x] `Raid1.CleanDegradedStartupLoadsFromActiveDevice` — M5 destructor persists superbitmap (`include_superbitmap=true`)
- [x] `Raid1.CleanDegradedStartupEmptySuperbitmapThrows` — M5 guard: zero superbitmap in clean-degraded state throws
- [ ] `GccThreadSanitize` CI — primary verification for M5 (local-buffer approach eliminates the TSan report on `superbitmap_reserved`)
- [ ] All existing RAID1 tests continue to pass

Closes #273, #277, #282.

### DIFF
--- a/src/raid/raid1/raid1.cpp
+++ b/src/raid/raid1/raid1.cpp
@@ -235,7 +235,8 @@ void Raid1Disk::__init_bitmap_and_degraded_route() {
         // Route reads to whichever physical slot is live
         _read_route_cache.store(a_is_missing ? read_route::DEVB : read_route::DEVA, std::memory_order_release);
         // Load bitmap from the live slot; don't bump age - we may get the original disk back via swap
-        _dirty_bitmap->load_from(*(a_is_missing ? _device_b : _device_a)->disk);
+        _dirty_bitmap->load_from(*(a_is_missing ? _device_b : _device_a)->disk,
+                                 static_cast< read_route >(_sb->fields.read_route) != read_route::EITHER);
     } else if (_device_a->new_device xor _device_b->new_device) {
         // Bump the bitmap age
         _sb->fields.bitmap.age = htobe64(be64toh(_sb->fields.bitmap.age) + 16);
@@ -254,7 +255,7 @@ void Raid1Disk::__init_bitmap_and_degraded_route() {
         auto const& active_dev = (route == read_route::DEVB) ? _device_b : _device_a;
         auto const& backup_dev = (route == read_route::DEVB) ? _device_a : _device_b;
         RLOGW("Raid1 is starting in degraded mode [uuid:{}]! Degraded device: {}", _str_uuid, *backup_dev->disk)
-        _dirty_bitmap->load_from(*active_dev->disk);
+        _dirty_bitmap->load_from(*active_dev->disk, true);
     } else if (0 == _sb->fields.clean_unmount) {
         RLOGW("Raid1 was not cleanly shutdown last time [uuid:{}]!", _str_uuid)
     }
@@ -301,8 +302,11 @@ Raid1Disk::~Raid1Disk() {
         RLOGI("Synchronized: [uuid: {}]", _str_uuid)
     }
     _sb->fields.clean_unmount = 0x1;
-    // Only update the superblock to clean devices
-    if (auto res = write_superblock(*state.active_dev->disk, _sb.get(), read_route::DEVB == state.route, state.route);
+    // Only update the superblock to clean devices. Pass include_superbitmap=true so the
+    // on-disk superbitmap reflects the current dirty state; load_from checks this on next
+    // startup when the array opens degraded (was_previously_degraded invariant).
+    if (auto res =
+            write_superblock(*state.active_dev->disk, _sb.get(), read_route::DEVB == state.route, state.route, true);
         !res) {
         if (state.is_degraded) {
             RLOGE("Failed to clear clean bit...full sync required upon next assembly [uuid:{}]", _str_uuid)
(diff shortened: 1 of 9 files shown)


src/raid/raid1/raid1.cpp
```diff
@@ -160,6 +160,20 @@ void Raid1Disk::__init_params() {
...
+        if (device->is_missing()) continue;
+        auto const dev_bytes = device->capacity();
+        if (dev_bytes < _reserved_size) {
+            RLOGE("Device {} is too small: {} bytes < reserved {} bytes [uuid:{}]", *device, dev_bytes, _reserved_size,
+                  _str_uuid)
+            throw std::runtime_error(
+                fmt::format("device too small: {} bytes < reserved {} bytes", dev_bytes, _reserved_size));
+        }
+    }
+
```
szmyd, line 174: Don't we already "pick" the smallest device as the overall array size? Do we need to look at both disks _again_?

szmyd (approved):
In general I am good with this change. Please make sure the final squash message is reflective of the final changes.

szmyd (changes requested):
version bump?

szmyd (approved):
LGTM

PR #287, opened 2026-05-28
chore(claude): add /bugs-hunt agentic command

Ilyin-Ilya:
Adds \`.claude/commands/bugs-hunt.md\`: a 3-phase agentic bug-hunting skill for this repo.

## How it works

**Phase 1 — Parallel discovery (5 agents)**
Five agents run concurrently, each scoped to one subsystem:

| Agent | Scope |
|---|---|
| 1 | \`src/raid/raid1/bitmap.cpp\`, \`bitmap.hpp\`, \`super_bitmap.cpp\`, \`super_bitmap.hpp\` |
| 2 | \`src/raid/raid1/raid1.cpp\` — state transitions (\`__become_*\`, \`__swap_device\`, \`async_iov\`) |
| 3 | \`src/raid/raid1/raid1_superblock.*\`, startup path (\`__init_*\`, \`__load_*\`) |
| 4 | \`src/target/ublkpp_tgt.cpp\`, \`src/target/ublkpp_tgt_impl.hpp\`, \`src/lib/ublk_disk.cpp\` |
| 5 | \`src/raid/raid0/raid0.cpp\`, \`src/driver/\`, \`src/metrics/\` |

All output files go in \`.claude/bugs-hunt/\`. Re-running overwrites previous results.

**Phase 2 — Independent verification**
One clean-context verification agent per candidate. No shared state with discovery. Each agent traces the exact trigger sequence through the code, identifies every CAS gate / lock / ordering constraint, and delivers a CONFIRM / REJECT / ESCALATE verdict.

**Phase 3 — Consolidated report**
Confirmed bugs, rejected false positives (with the specific guard that blocks them), and escalations (precise scenario questions for the user).

## Key principles

- **Million-dollar rule**: only report if 100% certain it is a real bug; uncertain findings become escalations, never bug claims. "No bugs found" is a valid and valuable outcome.
- **Power-loss safety**: every line can be the last — never assume line A+1 ran because line A did.
- **CAS gate analysis**: trace the full compare-exchange chain before flagging a race — concurrent execution is often structurally impossible.
- **Race direction**: a set-only race (\`fetch_or\`) goes in the safe direction; a clear race (\`fetch_and\`) is the data-loss direction. Classify accurately.
- **Known false positives**: RAID1-specific phantom patterns (destructor/IO-coroutine races, concurrent \`write_superblock\` callers) are pre-documented with invariant anchors so agents don't rediscover them on every run.

### DIFF
--- a/.claude/commands/bugs-hunt.md
+++ b/.claude/commands/bugs-hunt.md
@@ -1,11 +1,13 @@
 # Bug Hunt — ublkpp
 
+<!-- Last updated: 2026-05-31. KFP entries reflect code as of this date — re-verify before applying if significant time has passed. -->
+
 Systematic correctness audit using parallel discovery agents followed by independent
 verification. Each finding must survive a separate clean-context challenge before it
 is reported as real.
 
-**Scope**: implementation files only (`src/`). Public headers (`include/ublkpp/`) and
-interface-level contracts are out of scope for this hunt.
+**Scope**: implementation files only (`src/`), including internal headers under `src/`.
+Out of scope: public API headers (`include/ublkpp/`) and interface-level contracts.
 
 ---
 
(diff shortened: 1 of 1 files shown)


d2chau (approved):
LGTM. 

I think the only thing I'm considering is how to make it so this skill can check in any bugs so that other people can view them too. Especially if they chose to work on it

PR #288, opened 2026-05-29
fix(bitmap,target,raid0): Bitmap-B2 race, Target-B2 queue failure detection, RAID0-B5 discard propagation

Ilyin-Ilya:
## Summary

Three confirmed bugs from issue #284, addressed after independent verification:

- **Bitmap-B2** (`src/raid/raid1/bitmap.cpp`): `clean_region` had a TOCTOU race between `isal_zero_detect` (non-atomic SIMD scan) and `clear_bit` (atomic). A concurrent `dirty_region` could set bits and mark the superbitmap between those two calls, then `clear_bit` would erase that mark — leaving the superbitmap inconsistent (bit cleared but page has dirty data). Fixed with a double-check pattern guarded by `std::atomic_thread_fence(seq_cst)`: after `clear_bit`, re-read the page; if it has new dirty bits, restore the superbitmap bit.

- **Target-B2** (`src/target/ublkpp_tgt.cpp`): `ublksrv_queue_handler` failures (null `q` from `ublksrv_queue_init`) were silently swallowed — `start()` would proceed as if all queues initialized successfully. Fixed by passing an `std::atomic<int>* queue_failures` counter to each queue thread; threads increment it before `sem_post` on failure; `start()` checks the counter after all semaphores are collected and returns an error if any queue failed.

- **RAID0-B5** (`src/raid/raid0/raid0.cpp`): `Raid0Disk` did not propagate `max_discard_sectors` from child devices. A top-level discard could be dispatched at a size the child device rejects. Fixed: constructor computes the minimum `max_discard_sectors` across children, then scales by stripe count (the aggregate size a RAID0 can handle is `min_child × num_stripes`). Added `max_discard_sectors()` public accessor to `ublk_disk` since the field was only reachable via the protected `params()`.

Bugs evaluated but confirmed as **false positives** (not fixed): RAID1-B1, RAID1-B6, RAID1-B9, Bitmap-B3, Bitmap-B4, RAID1-B14, SB-B9. All relied on CAS gates, type-safe invariants, or sequencing guarantees that make the "concurrent" path structurally impossible.

## Test plan

- [ ] `Raid1CleanRegion/FullCleanClearsSuperbitmap` — new: full clean clears superbitmap, returns non-null clean page
- [ ] `Raid1CleanRegion/PartialCleanPreservesSuperbitmap` — new: partial clean leaves superbitmap set, returns null
- [ ] `Raid1CleanRegion/DirtyCleanDirtyCyclePreservesInvariant` — new: multi-round dirty→clean cycle stays consistent
- [ ] `Raid1CleanRegion/CleanAlreadyCleanPageIsNoOp` — new: cleaning unallocated page is safe
- [ ] `Raid0/MaxDiscardSectorsPropagated` — new: min-child × stripe-count propagation verified
- [ ] CI TSan job covers the concurrent Bitmap-B2 race window (deterministic unit test not feasible for that path)
- [ ] CI GccAddressSanitize, GccCoverage, ClangRelease jobs

### DIFF
--- a/include/ublkpp/lib/ublk_disk.hpp
+++ b/include/ublkpp/lib/ublk_disk.hpp
@@ -60,6 +60,7 @@ public:
     uint32_t max_tx() const noexcept;
     bool can_discard() const noexcept;
     uint32_t discard_granularity() const noexcept;
+    uint32_t max_discard_sectors() const noexcept;
     uint64_t capacity() const noexcept;
     bool direct_io() const noexcept { return _direct_io; }
     bool is_missing() const noexcept { return _is_missing; }
--- a/src/raid/raid1/bitmap.cpp
+++ b/src/raid/raid1/bitmap.cpp
@@ -328,8 +328,17 @@ std::tuple< Bitmap::word_t*, uint32_t, uint32_t > Bitmap::clean_region(uint64_t 
 
     // Check if page became completely clean, and update superbitmap if so
     if (0 == isal_zero_detect(page, k_page_size)) {
-        // Clear superbitmap bit for this now-clean page
         _super_bitmap.clear_bit(page_offset);
+        // Race guard: dirty_region may have called fetch_or (setting bits) and then set_bit
+        // (marking superbitmap) between our isal_zero_detect and clear_bit above. The fence
+        // ensures we observe any such concurrent store before re-reading the page. If the page
+        // has new dirty bits we restore the superbitmap bit; if dirty_region's set_bit runs
+        // after our fence it wins and sets the bit itself — either way the superbitmap is correct.
+        std::atomic_thread_fence(std::memory_order_seq_cst);
+        if (0 != isal_zero_detect(page, k_page_size)) {
+            _super_bitmap.set_bit(page_offset);
+            return std::make_tuple(nullptr, page_offset, sz);
+        }
         return std::make_tuple(_clean_page.get(), page_offset, sz);
     }
     return std::make_tuple(nullptr, page_offset, sz);
--- a/src/lib/ublk_disk.cpp
+++ b/src/lib/ublk_disk.cpp
@@ -49,6 +49,7 @@ uint32_t ublk_disk::physical_block_size() const noexcept { return 1 << _params->
 uint32_t ublk_disk::max_tx() const noexcept { return _params->basic.max_sectors << SECTOR_SHIFT; }
 bool ublk_disk::can_discard() const noexcept { return _params->types & UBLK_PARAM_TYPE_DISCARD; }
 uint32_t ublk_disk::discard_granularity() const noexcept { return _params->discard.discard_granularity; }
+uint32_t ublk_disk::max_discard_sectors() const noexcept { return _params->discard.max_discard_sectors; }
 uint64_t ublk_disk::capacity() const noexcept { return _params->basic.dev_sectors << SECTOR_SHIFT; }
 
 namespace {
--- a/src/raid/raid0/raid0.cpp
+++ b/src/raid/raid0/raid0.cpp
@@ -144,8 +144,16 @@ Raid0Disk::Raid0Disk(boost::uuids::uuid const& uuid, uint32_t const stripe_size_
     // Align size to max_sector size
     our_params.basic.dev_sectors -= (our_params.basic.dev_sectors % our_params.basic.max_sectors);
 
-    if (can_discard())
+    if (can_discard()) {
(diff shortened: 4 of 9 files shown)


src/raid/raid1/bitmap.cpp
```diff
@@ -328,8 +328,17 @@ std::tuple< Bitmap::word_t*, uint32_t, uint32_t > Bitmap::clean_region(uint64_t
...
     if (0 == isal_zero_detect(page, k_page_size)) {
-        // Clear superbitmap bit for this now-clean page
         _super_bitmap.clear_bit(page_offset);
+        // Race guard: dirty_region may have called fetch_or (setting bits) and then set_bit
+        // (marking superbitmap) between our isal_zero_detect and clear_bit above. The fence
+        // ensures we observe any such concurrent store before re-reading the page. If the page
+        // has new dirty bits we restore the superbitmap bit; if dirty_region's set_bit runs
+        // after our fence it wins and sets the bit itself — either way the superbitmap is correct.
+        std::atomic_thread_fence(std::memory_order_seq_cst);
+        if (0 != isal_zero_detect(page, k_page_size)) {
```
Ilyin-Ilya, line 338: Yes, the fix is needed on x86 too. My earlier answer was only about whether the C++ abstract machine's "non-atomic read of atomically-written memory" UB causes practical harm on x86 — it doesn't. But that's
  a separate question from whether the TOCTOU race itself exists.

  The race is about thread interleaving, not memory visibility.

  On x86, threads still interleave freely between operations. The bug scenario is:

  Thread C (clean_region)          Thread D (dirty_region)
  isal_zero_detect → 0
                                   fetch_or → bits set in page
                                   set_bit  → superbitmap marked
  clear_bit → superbitmap cleared

  x86 TSO says stores are immediately visible — it says nothing about preventing this ordering. clear_bit absolutely can erase set_bit's write on x86, because they're separate atomic RMWs on the same byte with
   no exclusion between them.

  The fix is also needed on x86 for the same reason: the mfence ensures thread C's second isal_zero_detect sees thread D's fetch_or if it happened before the fence, and thread D's own set_bit handles the case
  where fetch_or happens after. Neither path depends on a weak memory model — they depend on closing the ordering gap between the two operations in thread C.

  The memory model only affected whether I could call the non-atomic scan "UB." The race itself is architecture-independent.

include/ublkpp/lib/ublk_disk.hpp
```diff
@@ -60,6 +60,7 @@ class ublk_disk {
     uint32_t max_tx() const noexcept;
     bool can_discard() const noexcept;
     uint32_t discard_granularity() const noexcept;
+    uint32_t max_discard_sectors() const noexcept;
```
Ilyin-Ilya, line 63: How discard works in ublk

  When the kernel wants to discard a range, it issues a UBLK_IO_OP_DISCARD to the RAID0 device. The kernel caps each request at the device's advertised max_discard_sectors. RAID0 then fans that out to child
  devices via merged_subcmds, and each child calls its own async_iov (which ultimately does a BLKDISCARD ioctl or fallocate(FALLOC_FL_PUNCH_HOLE) on the underlying file/device).

  The bug

  Before the fix, Raid0Disk left max_discard_sectors at the ublk_disk default: UINT_MAX >> SECTOR_SHIFT — effectively unlimited.

  So the kernel sees "this RAID0 can handle arbitrarily large discards" and sends one big request. Say it sends 2 GB:

  Kernel: DISCARD 2 GB → RAID0 (2 stripes)
    RAID0 → merged_subcmds → child A: DISCARD 1 GB
                             child B: DISCARD 1 GB
    child A's max_discard_sectors: 512 MB
    child A rejects with EINVAL

  The child's underlying block device enforces its own limit. RAID0 never checked it — it just forwarded whatever the kernel gave it divided by the stripe count.

  Why the formula min_child × stripe_count is correct

  A RAID0 discard of N total sectors distributes N / stripe_count sectors to each child (stripe-aligned). So the largest N that no child can reject is:

  N / stripe_count ≤ min(child.max_discard_sectors)
  N ≤ min_child × stripe_count

  The fix sets max_discard_sectors = min_child × stripe_count. The kernel now splits large discards into chunks of at most that size. When RAID0 fans each chunk out, every child receives at most min_child
  sectors — within its own limit.

  Why it was invisible

  The default child max_discard_sectors is also UINT_MAX >> 9 (set in ublk_disk.cpp:31). If all children are FSDisk backed by a regular file, the fallocate call has no practical sector limit, so the bug never
  fired. It only bites with real block devices that enforce a smaller max_discard_sectors at the hardware level — SSDs in particular tend to have this limit.

PR #289, opened 2026-05-29
fix(bitmap,target,raid0): Bitmap-B2 race, Target-B2 queue failure detection, RAID0-B5 discard propagation

Ilyin-Ilya:
## Summary

Three confirmed bugs from issue #284, addressed after independent verification:

- **Bitmap-B2** (`src/raid/raid1/bitmap.cpp`): `clean_region` had a TOCTOU race between `isal_zero_detect` (non-atomic SIMD scan) and `clear_bit` (atomic). A concurrent `dirty_region` could set bits and mark the superbitmap between those two calls, then `clear_bit` would erase that mark — leaving the superbitmap inconsistent (bit cleared but page has dirty data). Fixed with a double-check pattern: after `clear_bit`, re-scan the page with `memory_order_acquire` loads. `dirty_region`'s `fetch_or` is upgraded to `memory_order_release` so the acquire loads in the double-check synchronize with it — if any word is non-zero, the concurrent write is visible and the superbitmap bit is restored. This is a proper release/acquire happens-before pair: portable to ARM/POWER and compatible with TSan (unlike `atomic_thread_fence`). On x86, release RMWs compile to the same `LOCK` instruction as relaxed RMWs, so there is no performance regression.

  `is_dirty` no longer uses the superbitmap as a fast-path skip for clean pages. Both `dirty_region`'s write→set_bit gap and `clean_region`'s clear_bit→double-check-restore window can transiently show the superbitmap as clean on a dirty page; with the fast-path in place, a degraded-mode read would route to the stale backup. Reading page bits directly (via `_page_map`) avoids this entirely. `next_dirty()` and `dirty_pages()` still use the superbitmap and are unaffected.

- **Target-B2** (`src/target/ublkpp_tgt.cpp`): `ublksrv_queue_handler` failures (null `q` from `ublksrv_queue_init`) were silently swallowed — `start()` would proceed as if all queues initialized successfully. Fixed by giving each queue thread its own `int` slot in a pre-allocated vector; the thread writes 0 on failure before `sem_post`. `start()` checks all slots after the semaphore loop (the semaphore provides the necessary happens-before) and tears down surviving threads via `destroy()` if any slot is 0. No atomic needed: each thread writes to a distinct index, eliminating concurrent writes to the same location.

- **RAID0-B5** (`src/raid/raid0/raid0.cpp`): `Raid0Disk` did not propagate `max_discard_sectors` from child devices. A top-level discard could be dispatched at a size the child device rejects. Fixed: constructor computes the minimum `max_discard_sectors` across children, then scales by stripe count (the aggregate size a RAID0 can handle is `min_child × num_stripes`). Added `max_discard_sectors()` public accessor to `ublk_disk` since the field was only reachable via the protected `params()`.

Bugs evaluated but confirmed as **false positives** (not fixed): RAID1-B1, RAID1-B6, RAID1-B9, Bitmap-B3, Bitmap-B4, RAID1-B14, SB-B9. All relied on CAS gates, type-safe invariants, or sequencing guarantees that make the "concurrent" path structurally impossible.

## Test plan

- [ ] `Raid1CleanRegion/FullCleanClearsSuperbitmap` — new: full clean clears superbitmap, returns non-null clean page
- [ ] `Raid1CleanRegion/PartialCleanPreservesSuperbitmap` — new: partial clean leaves superbitmap set, returns null
- [ ] `Raid1CleanRegion/DirtyCleanDirtyCyclePreservesInvariant` — new: multi-round dirty→clean cycle stays consistent
- [ ] `Raid1CleanRegion/CleanAlreadyCleanPageIsNoOp` — new: cleaning unallocated page is safe
- [ ] `Raid1/IsDirtyReflectsPageBitsAfterSuperbitmapCleared` — new: dirty/clean/redirty cycle verified via direct page-bit read
- [ ] `Raid0/MaxDiscardSectorsPropagated` — new: min-child × stripe-count propagation verified (2 stripes)
- [ ] `Raid0/MaxDiscardSectorsPropagatedThreeStripes` — new: min-child selection and multiply verified (3 stripes)
- [ ] CI TSan job covers the concurrent Bitmap-B2 race window (deterministic unit test not feasible for that path)
- [ ] CI GccAddressSanitize, GccCoverage, ClangRelease jobs


src/raid/raid1/bitmap.cpp
```diff
@@ -328,8 +335,24 @@ std::tuple< Bitmap::word_t*, uint32_t, uint32_t > Bitmap::clean_region(uint64_t
...
+        // Race guard: dirty_region may have called fetch_or (setting bits) and then set_bit
+        // (marking superbitmap) between our isal_zero_detect and clear_bit above. Re-scan the
+        // page with acquire loads to synchronize with dirty_region's release fetch_or: if any
+        // word is non-zero the concurrent write is now visible, and we restore the superbitmap
+        // bit. If dirty_region's set_bit runs after our clear_bit it wins and sets the bit
+        // itself — either way the superbitmap is correct.
+        // Using acquire loads (rather than atomic_thread_fence) keeps TSan happy and makes the
+        // fix portable to ARM/POWER: the release/acquire pair creates the required happens-before.
+        auto const nwords = k_page_size / sizeof(word_t);
+        auto* words = page;
```
Ilyin-Ilya, line 348: this is because tsan does not work with atomic_thread_fence, not supported. 
The bonus is that this works on arm.

szmyd (approved):
This looks correct to me. :+1:

PR #293, opened 2026-05-31
fix(raid1): close __become_clean / backup-fail dirty_region race

Ilyin-Ilya:
## Summary

Fixes the P0 race where `__become_clean()` could transition the route to `EITHER` while dirty bits remained in the bitmap, with no resync to process them.

## Root cause

When the resync task observes `dirty_pages()==0` and calls `__become_clean()`, a concurrent write whose backup leg fails can call `dirty_region()` in the gap between the observation and the CAS. The original code left those dirty bits stranded with route=EITHER and no resync running — subsequent reads skipped the `is_dirty` guard (only active in degraded mode) and could be served from the stale backup.

## Three-part fix

**1. Backup-fail paths (async_iov + sync_iov):** Remove the `if (!state.is_degraded)` guard; call `__become_degraded(false, &state)` unconditionally. `__become_degraded` CAS-es from EITHER: if the array is still degraded the CAS fails with `old==new_route` → no-op; if the array drifted to clean while the write was in-flight the CAS succeeds and re-degrades the correct leg before the ACK. Must use the captured `&state` — not a live re-capture — because `failed_is_active` is relative to the captured device assignment.

**2. `__become_clean()` returns `bool`:** CAS to EITHER, re-check `dirty_pages()`, and if dirty bits exist reverse the CAS and return `false` instead of calling `toggle_resync(true)` (which is a no-op from inside a still-ACTIVE task). Also adds a post-write route check (H1): after writing the clean superblocks, if `__become_degraded` CAS-ed EITHER→degraded while the writes were in-flight, overwrite the superblocks with the current degraded route to prevent EITHER from persisting as the final on-disk state on crash.

**3. `_start()` loops on `complete()`:** The resync loop re-runs `__run()` whenever `complete()` returns `false` or `dirty_pages()>0` after a sweep, staying ACTIVE throughout. This is the liveness half — without it, the task exits to IDLE with dirty bits unsynced and no mechanism to re-launch (toggle_resync from inside ACTIVE is a no-op).

## Additional fixes

- `__become_clean` initial CAS failure: return `old_route == EITHER` instead of unconditional `true`, so a swap race (old_route != EITHER) correctly returns false and keeps the loop going.
- `SuperBitmap::set_bit`: upgraded to `memory_order_release` (was relaxed) so page-word writes preceding `set_bit()` are visible to `next_dirty()`/`clean_region()` acquire-load consumers.
- `is_dirty` word loads: upgraded to `memory_order_acquire` (was relaxed) to pair with `dirty_region`'s release `fetch_or` on ARM/POWER.

## Test plan

- [ ] `BecomeCleanReturnsFalseLoopsAndDrains` — deterministic: complete() returns false first, true second; asserts dirty_pages()==0 and call_count==2
- [ ] `IsDirtyPageBitsSetSuperbitmapClearedManually` — is_dirty reads page bits even with superbitmap cleared
- [ ] CI TSan job covers the concurrent race windows non-deterministically
- [ ] CI GccAddressSanitize, GccCoverage, ClangRelease


src/raid/raid1/raid1.cpp
```diff
@@ -619,12 +619,35 @@ void Raid1Disk::__become_clean() {
...
+    auto old_route = state.route;
+    if (!_read_route_cache.compare_exchange_strong(old_route, read_route::EITHER)) return;
+
+    // Double-check: a concurrent backup-fail dirty_region() may have fired between the
+    // resync task's dirty_pages()==0 observation and the CAS above. Reverse the
+    // transition so those bits are not stranded with no resync to process them.
+    if (_dirty_bitmap->dirty_pages() > 0) {
+        auto either_route = read_route::EITHER;
+        if (_read_route_cache.compare_exchange_strong(either_route, state.route)) {
+            RLOGW("Resync complete but found unsynced dirty bits — re-launching resync [uuid:{}]", _str_uuid)
```
Ilyin-Ilya: _read_route_cache.compare_exchange_strong(old_route, read_route::EITHER);
  // ← WINDOW OPENS: is_degraded = false, reads round-robin to A or B
     
  if (_dirty_bitmap->dirty_pages() > 0) {
      auto either_route = read_route::EITHER;
      _read_route_cache.compare_exchange_strong(either_route, state.route);
  // ← WINDOW CLOSES: route back to DEVA, reads go to active only

  The window contains exactly:
  1. The dirty_pages() call — a linear scan of the superbitmap (up to 256 bytes for a 2 TB volume) with acquire loads
  2. The branch evaluation
  3. Two instructions to set up the reverse CAS
  4. The reverse CAS itself
  
  On a cache-warm superbitmap this is 50–200 ns on a modern CPU. The superbitmap is small (256 bytes) and likely L1-hot since the resync task has been reading it continuously.

  But the window has two conditions that must both be true for a stale read to occur:

  First, the dirty_region() from the backup-fail must have fired during the gap between the resync task's dirty_pages() == 0 observation and the CAS — that's already a narrow timing requirement. The gap is:
  resync task checks bitmap at line 125, calls complete(), which calls __become_clean(), which executes several lines before the CAS. Maybe 200–500 ns of function call overhead.

  Second, a read for the specific address range that was dirtied must arrive during the 50–200 ns window and be routed to the backup device (50% probability in EITHER mode).

  So the actual exposure is the intersection: a write whose backup fails at exactly the right moment, AND a concurrent read for the same address range within ~200 ns. In practice this is extraordinarily
  unlikely — but it is a real window, not zero.

szmyd (dismissed):
Nice catch on the root cause, and the CAS-before-superblock inversion is a genuine improvement (crash safety + it closes the most common interleaving). The `memory_order` upgrade is fine too. One blocking issue and one framing concern.

## 🔴 Blocking: the reverse-path resync re-launch is a no-op

`__become_clean()` runs *inside* the resync task, while that task's atomic state is still `ACTIVE`:

- `raid1_resync_task.cpp:125` calls `complete()` with state `ACTIVE`
- the task only CASes `ACTIVE -> IDLE` *after* `complete()` returns (`:129-131`)

So in the new reverse branch:

```cpp
if (_resync_enabled.load(std::memory_order_relaxed)) toggle_resync(true);
```

`toggle_resync(true)` -> `launch()` -> `__transition_to(IDLE, IDLE, ...)` hits the `ACTIVE -> EARLY_EXIT` handler case and returns "already running" without spawning a thread. Then `complete()` returns, `_start()` sets `IDLE`, and the thread exits.

Net: the route is correctly reverted to degraded (reads stay safe), but **no resync runs and none gets re-launched**. A later backup failure in the same direction doesn't help either: `__become_degraded` finds the route already non-`EITHER`, the CAS fails with `old == new_route`, returns `0`, and never calls `toggle_resync`. The array is stuck degraded with a permanently-stale backup region until restart. If the active leg dies in the meantime, that un-resynced region is lost (durability hole).

**Fix:** the re-launch can't come from within the still-`ACTIVE` task. It needs to come back as feedback to the resync task telling it to keep going. Make `complete` return a bool ("became clean") and have `_start()` loop on it instead of unconditionally going idle:

```cpp
// __become_clean: CAS to EITHER; re-check dirty_pages():
//   if >0 -> reverse CAS to state.route and return false   (NO toggle_resync here)
//   else  -> write clean superblocks and return true

// _start():
while (true) {
    cur_state = __run(clean_mirror, dirty_mirror, &iov);
    if (resync_state::STOPPING == cur_state) break;
    DEBUG_ASSERT_EQ(resync_state::ACTIVE, cur_state, "...");
    if (0 != _dirty_bitmap->dirty_pages()) continue;  // interrupted
    if (complete()) break;                             // became clean -> done
    // complete() reversed route to degraded; dirty bits remain -> loop,
    // __run() re-reads dirty_pages() and resyncs the newly-dirtied region
}
```

`__run()` already re-reads `dirty_pages()` at its top (`:211`), so re-entering it picks up the concurrently-dirtied region and resyncs to convergence. State stays `ACTIVE` throughout, region tracker intact.

## 🟠 Framing: residual window is wider than "nanoseconds" and not always self-correcting

The forward path writes two synchronous superblocks (`write_superblock` -> `sync_iov`, real I/O) *after* the re-check. An in-flight write that captured `state.is_degraded == true` before the CAS, whose backup fails *during those SB writes*, takes the `async_iov` path:

```cpp
if (backup_res < 0) { _dirty_bitmap->dirty_region(addr,len); if (!state.is_degraded) __become_degraded(false,...); }
```

`state.is_degraded` is the captured (stale) `true`, so it dirties but does NOT re-degrade or spawn resync (it trusts the resync task, which is exiting). `__become_clean` already passed its re-check, so it finishes with route `EITHER` + dirty bits + no resync -> stale reads, permanently. So that window is SB-write latency (ms on iSCSI), not nanoseconds, and this class is not self-correcting. The re-check still strictly shrinks the original window, so it's a net win - just please soften the "permanently broken -> transiently exposed, then self-correcting" wording to acknowledge the in-flight-degraded-write class. (The "keep going" loop above does not close this one either; fully closing it needs the write/CAS mutex you mention, or routing the stale-read guard off something other than `is_degraded`.)

## 🟡 Minor

- The `memory_order` upgrade to `release` is correct and worth keeping, but the rationale isn't quite the mechanism: coherence already guarantees a later `next_set_bit` acquire-load observes a completed `fetch_or` even under `relaxed`, so the `dirty_pages()` double-check would see the bit regardless. The real value of `release/acquire` is making the page-word writes that precede `set_bit()` visible to `next_dirty()`/`clean_region()` consumers. Fine to keep, just adjust the note.
- Version/CHANGELOG collide with #294 (also `0.32.7`, also adds a `## [0.32.7]` section). Whichever merges second conflicts on `conanfile.py` + `CHANGELOG.md` - let's coordinate a shared `0.32.7` or bump one to `0.32.8`.

The read-safety half of this is solid; it's the resync-resume half (#1) I'd block on.

szmyd (changes requested):
Consolidated review (supersedes my earlier thread). The CAS-before-superblock inversion and the `memory_order` fix are good; here's the full design needed to actually close the race. Credit @szmyd for the reframing and for catching the post-ACK hole.

## The bug we must close: a post-ACK stale read

```
IO THREAD                      |  RESYNC THREAD
                               |   dirty_pages() == 0  -> __become_clean()
write captures DEGRADED        |
  backup_write = true          |
  backup write FAILS           |
  dirty_region()               |   (route still degraded here, so the write's
  __become_degraded -> no-op    |    __become_degraded CAS(EITHER->..) fails: already degraded)
  [ACK]                        |
                               |   CAS(route -> EITHER)
read captures CLEAN (EITHER)   |
  ~returns STALE backup~       |   dirty_pages() != 0  -> revert CAS -> DEVA
```

The write set a dirty bit and ACK'd while the route was degraded (so it was safe at that instant and couldn't self-heal -- its `__become_degraded` was a correct no-op). Then `__become_clean` does **check-then-CAS**: it flips the route to `EITHER` with the bit already set, and only re-checks `dirty_pages()` *after* the CAS. In that `[CAS, revert]` window an `EITHER`-mode read skips the `is_dirty` guard and serves the stale backup -- a read of an already-acknowledged LBA. The revert closes it going forward, but the read already escaped.

**Root cause / invariant:** `EITHER`-mode reads skip `is_dirty`, so the invariant is "whenever a lock-free reader observes `route == EITHER`, both legs are identical." `check-then-CAS` (and a plain route CAS, which a `dirty_region` doesn't perturb) has an instant that violates it. No amount of re-checking *after* the CAS removes that instant.

## The fix (4 parts)

**Part 1 - failure paths degrade off the CAPTURED state, unconditionally.** The backup-fail path's `if (!state.is_degraded)` guard suppresses the redegrade when captured-degraded; drop it so it matches the active-fail path (which already calls `__become_degraded` unconditionally):

```diff
     if (backup_res < 0) {
         _dirty_bitmap->dirty_region(addr, len);
-        if (!state.is_degraded) {
-            if (auto d = __become_degraded(false, &state); !d) co_return -EIO;
-        }
+        if (auto d = __become_degraded(false, &state); !d) co_return -EIO;
     }
```

Use the **captured** `&state`, not a fresh capture: `__become_degraded` identifies the failed leg from `(failed_is_active, route)`, and `failed_is_active` is relative to the capture. A live route would mis-identify the leg when the captured route was `DEVB`. `__become_degraded` CAS-es from `EITHER`, so this is a no-op when already degraded and re-degrades (correct leg) when the array drifted clean -- handling writes that capture `EITHER` or straddle the transition, before they ACK. (`sync_iov` is the same one-line guard removal; active-fail needs no change.)

**Part 2 - `__become_clean()` returns a "keep going" signal**, and drop the `toggle_resync(true)` from the revert path (it is a no-op from inside the still-`ACTIVE` resync task):

```cpp
bool __become_clean() {
    // ... under the part-4 lock:
    if (_dirty_bitmap->dirty_pages() != 0) return false;   // bits remain -> stay degraded
    CAS(route -> EITHER);
    // unlock; write clean superblocks
    return true;
}
```

**Part 3 - `_start()` loops on the signal** (liveness; the resync re-launch cannot come from inside the `ACTIVE` task):

```cpp
for (;;) {
    cur_state = __run(clean_mirror, dirty_mirror, &iov);
    if (resync_state::ACTIVE != cur_state) break;          // STOPPING
    if (0 != _dirty_bitmap->dirty_pages()) continue;       // defensive
    if (complete()) break;                                 // became clean
    // else complete() found late bits -> loop, __run() drains them
}
```
(`complete`: `function<void()>` -> `function<bool()>`; callback `[this]{ return __become_clean(); }`.)

**Part 4 - a cold-path transition lock** makes `become_clean`'s check+CAS atomic against the failure-path `dirty_region`:
- `__become_clean`: `lock -> if (dirty_pages()) return false; CAS(route -> EITHER) -> unlock` (SB writes can stay outside).
- the leg-**failure** blocks: `lock -> dirty_region(); __become_degraded(..., &state) -> unlock`.

With this, `check-then-CAS` is atomic w.r.t. new dirty bits: a failing write either lands before become_clean (sees `dirty_pages != 0`, no CAS) or after (captures/sees `EITHER`, re-degrades before ACK). No `EITHER`-with-dirty-bit instant is ever observable to a lock-free reader.

## The lock does NOT go in the fast `dirty_region` path

There are three `dirty_region` sites in the write path; only two can race `become_clean`:

1. **active fails** (`active_res < 0`, `backup_write == true`) - newly dirties a clean region. **Needs the lock.**
2. **`!backup_write`** (the warm degraded re-dirty) - reached only when `is_degraded && (backup unavail || is_dirty)`. So either the region is **already dirty** (`dirty_pages > 0` -> become_clean isn't transitioning) or the **backup is unavail** (resync's dirty_mirror is down -> resync blocked -> `dirty_pages` never hits 0). It cannot coincide with the transition, and only re-marks an already-set bit. **No lock.**
3. **backup fails** (`backup_res < 0`, `backup_write == true`) - newly dirties a clean region. **Needs the lock.**

So the lock is only on **actual leg failures** (cold by definition) plus the resync-completion path. Successful writes and reads are entirely lock-free; contention is essentially nil, and everything inside the locked sections is synchronous (no `co_await` under the lock). Please don't wrap all of `dirty_region` - site #2 must stay lock-free.

## Net

Part 1 (safety for self-healing writes) + Part 3 (liveness) + Part 4 (atomic transition, closing the post-ACK window) together restore the invariant at every instant; Part 2 wires the signal. `__become_clean` ends up needing atomicity only against the cold failure path, not the hot write path.

Happy to pair on the patch and a TSan repro.

szmyd (changes requested):
The four-part design is correctly implemented and the CI confirms it. Two items before merge.

## `on_idle_dirty` can be eliminated

The callback + self-join detection is working around the fact that `_start()` exits before the ACTIVE->IDLE CAS
is visible to other threads. The simpler fix is to include the CAS inside the loop and retry if dirty bits
appear in the transition gap:

```cpp
while (true) {
    cur_state = __run(clean_mirror, dirty_mirror, &iov);
    if (resync_state::STOPPING == cur_state) break;
    if (0 != _dirty_bitmap->dirty_pages()) continue;
    if (!complete()) continue;
    if (0 != _dirty_bitmap->dirty_pages()) continue;

    // Attempt IDLE transition; recheck for bits that landed in the gap.
    auto active = resync_state::ACTIVE;
    __cas_state(active, resync_state::IDLE);
    if (0 == _dirty_bitmap->dirty_pages()) break;

    // Bits appeared between the loop check and the CAS. Try to reclaim ACTIVE.
    auto idle = resync_state::IDLE;
    if (!__cas_state(idle, resync_state::ACTIVE)) break; // another launch() won — they handle it
}
free(iov.iov_base);
// metrics...
// no on_idle_dirty call needed
```

This drops the `on_idle_dirty` parameter from `_start()`/`launch()`, the self-join `get_id()` check, and the
`detach()` branch — three mechanisms that exist solely to work around the gap. The CAS-back fails gracefully
when a concurrent `launch()` beats us to it.

## Minor: `is_missing()` asymmetry inside the lock

H1 guards the backup superblock write:

```cpp
if (!live_state.backup_dev->disk->is_missing()) {
    write_superblock(*live_state.backup_dev->disk, ...)
}
```

The EITHER writes inside the lock do not:

```cpp
write_superblock(*state.backup_dev->disk, ...)  // no is_missing() check
```

Not a correctness issue (the failed write is logged, pick_superblock recovers by age, missing backup is caught
at boot), but inconsistent. Worth a one-liner guard for symmetry.

szmyd (approved):
Both items addressed. The loop-with-CAS implementation is clean - the inner spin for spurious weak-CAS failures on ARM/POWER is a good touch, and `cur_state = __load_state()` correctly normalizes state for the STOPPING check after the multi-exit loop. The `is_missing()` guard is in place inside the lock. Good to merge.

PR #294, opened 2026-06-01
fix(raid0): floor legs to whole stripes so capacity isn't over-reported

szmyd:
Array capacity was (min_leg_capacity - stripe_size) * leg_count, which assumes every leg holds a whole number of stripes. When a leg capacity is not a stripe multiple, the partial trailing stripe was carried into the per-leg term and then multiplied by leg_count, over-reporting the device by (leg_capacity % stripe_size) * leg_count. Reads into that phantom tail (top-of-device GPT / partition-table scans) mapped to a per-leg offset past the end of the backing device and returned EIO.

Latent until raid1 SuperBlock v2 dropped the 512 KiB user-data alignment that had kept leg capacities stripe-aligned; surfaced on a 50-leg RAID10 of 3 TiB legs over-reporting by ~3 MiB. Fixed by flooring the smallest leg to a whole number of stripes before reserving the head stripe and striping.

Adds regression test Raid0.PartialStripeLegNotOverReported; bumps 0.32.7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.32.7] - 2026-06-01
+
+### Fixed
+
+- **H3 (raid0)**: the array capacity was computed as `(min_leg_capacity - stripe_size) * leg_count`, which assumes every leg holds a whole number of stripes. When a leg's capacity is not a multiple of `stripe_size`, the partial trailing stripe (`leg_capacity % stripe_size`) was carried into the per-leg term and then multiplied by the leg count, over-reporting the device size by `(leg_capacity % stripe_size) * leg_count`. Reads into that phantom tail (e.g. backup-GPT / partition-table scans at the top of the device) mapped to a per-leg offset past the end of the backing device and returned EIO. The bug was latent because leg capacities happened to be stripe-aligned; raid1 SuperBlock v2 dropped the 512 KiB user-data alignment that had masked it, surfacing top-of-device read failures on RAID10 arrays (observed on a 50-leg array of 3 TiB legs over-reporting by ~3 MiB). Fixed by flooring the smallest leg to a whole number of stripes before reserving the head stripe and striping.
+
 ## [0.32.6] - 2026-05-26
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


Ilyin-Ilya (commented):
The fix itself: correct

  The one-line insertion is in the right place and does the right thing. our_params.basic.dev_sectors holds the per-leg minimum in 512-byte sectors at that point (accumulated via std::min in the device loop).
  Flooring it to a stripe multiple before subtracting the superblock stripe and multiplying by leg count is exactly correct. All edge cases are handled:

  - Already stripe-aligned → H3 is a no-op ✓
  - dev_sectors < stripe_size_sectors → H3 zeroes it; H2 throws ✓
  - dev_sectors == stripe_size_sectors → H3 no-op; H2 throws ✓
  - Superblock-corrected _stripe_size → H3 uses _stripe_size (already corrected by line 120–124) ✓

  The "~3 MiB" claim in the description is verified: a 3 TiB RAID1 leg under SB v2 has 61,440 bytes of trailing partial stripe; 61,440 × 50 legs = 2.93 MiB ≈ 3 MiB.

  ---
  The test: two critical failures
  
  Failure 1 — wrong expected value. safe_max does not account for the final alignment step at line 145:

  our_params.basic.dev_sectors -= (our_params.basic.dev_sectors % our_params.basic.max_sectors);

  With the default max_io = 512 KiB and n=8 legs, max_sectors = 8192 sectors = 4 MiB. After H3 + subtract + multiply, dev_sectors = 16,775,168. The alignment trims 6,144 sectors, giving capacity() = 
  8,585,740,288 bytes. But safe_max = 8,588,886,016 — off by 3,145,728 bytes. The assertion fails against the post-fix code.

  Failure 2 — the fix is invisible to the test. With these exact parameters the alignment step happens to absorb the entire H3 correction:

  pre-fix:  16,776,192 - (16,776,192 % 8192) = 16,769,024 sectors → 8,585,740,288 bytes
  post-fix: 16,775,168 - (16,775,168 % 8192) = 16,769,024 sectors → 8,585,740,288 bytes

  Pre-fix and post-fix produce identical capacity. The test cannot detect the bug with these parameters. Verified independently:

  pre==post: True
  post==safe_max: False  (off by 3,145,728 bytes)

  To write a passing test that actually proves the fix: choose parameters where stride_width divides max_sectors (so the alignment is a no-op), e.g. max_io = stripe (set TestParams{.capacity = leg_cap, .max_io
  = stripe}), which makes max_sectors = stride_width = n × stripe_sectors and the alignment step becomes a no-op. Then safe_max is the correct expected value.

  ---

Ilyin-Ilya (approved):

PR #297, opened 2026-06-03
ublkpp: consume stdexec via conan (drop FetchContent)

szmyd:
sisl now ships stdexec as a conan package, so drop ublkpp's stdexec FetchContent block and resolve it through the conan graph: find_package(stdexec) + link stdexec::stdexec (which carries the include dirs, the _PSTL_PAR_BACKEND_SERIAL define, and the gcc warning flags). Bump the sisl floor to ^14.4 (the sisl/async headers + the stdexec dependency landed there). No other conanfile change -- stdexec comes in transitively via sisl.


Ilyin-Ilya (approved):
Minor — two public headers removed without CHANGELOG entry.

  include/ublkpp/lib/disk_task.hpp and include/ublkpp/lib/ublk_params.hpp are deleted. External ublk_disk subclasses that include either directly will fail to compile after the upgrade. ublk_params.hpp's
  content is now inlined into ublk_disk.hpp (fine, users should include that instead). disk_task.hpp is replaced by the type aliases in ublk_disk.hpp (also fine). Worth a CHANGELOG note under ### Changed so
  downstream adopters know to drop the direct includes.
  
  Nit — verbose comment on the #include <ublk_cmd.h> line in ublk_disk.hpp.
  The multi-line comment describes the kernel UAPI types but lives on an include directive inside a header — the old ublk_params.hpp comment was the appropriate location for that prose. Could move to a block
  above the include or trim it.

PR #298, opened 2026-06-03
feat(target): add begin_shutdown() for clean backing-store flush on SIGTERM

Ilyin-Ilya:
## Background

On SIGTERM, the daemon needs to flush the RAID-1 dirty bitmap and write `clean_unmount=1` to the superblock before the process exits — otherwise the next mount sees a dirty bitmap and must resync. The ublk kernel device state (QUIESCED vs DEAD) is a secondary concern.

The previous approach attempted to implement `UBLK_S_DEV_QUIESCED` in the destructor by calling `ctrl_deinit` without `stop_dev`. That approach does not work:
- `UBLK_S_DEV_QUIESCED` requires Linux 6.8; the target environment is Linux 6.6.
- `ctrl_deinit` closes the control fd, not the per-queue char device fds — `UBLK_IO_RES_ABORT` is never delivered to queue threads, so `join()` in the destructor would deadlock.
- On Linux 6.6, the only in-process mechanism to make queue threads exit is `stop_dev`, which transitions to `UBLK_S_DEV_DEAD` and drops `/dev/ublkbN`.

## What this PR does instead

Adds `ublkpp_tgt::begin_shutdown()` and `ublkpp_tgt::wait_for_drain()` — new APIs for backing-store drain before process exit.

```
begin_shutdown()          → stores _shutting_down atomic
__handle_io_async gate    → rejects all new I/O with EAGAIN (before device)
last in-flight op drains  → CAS on _device_reset_done → device.reset() fires once
                            (RAID-1: flushes dirty bitmap, writes clean_unmount=1)
                         → _drain_complete.notify_all() unblocks wait_for_drain()
process exits             → OS reclaims ublk fds; kernel handles device state
```

The destructor is reverted to an empty stub — ublk stack cleanup is left to the OS.

## Shutdown path comparison

| Path | Trigger | Backing store |
|---|---|---|
| `begin_shutdown()` + `wait_for_drain()` + process exit | SIGTERM handler | `clean_unmount=1` written |
| `ublkpp_tgt::remove()` → `destroy()` | explicit removal | flushed via `device.reset()` in `destroy()` |
| crash (no `begin_shutdown()`) | OOM / kill -9 | dirty — resync on next mount |

## Details

- `begin_shutdown()` stores a single `std::atomic<bool>` — safe from a SIGTERM handler.
- `__handle_io_async` increments the in-flight counter before the shutdown gate, then decrements and rejects with EAGAIN if `_shutting_down` is set. The increment-before-gate ordering prevents a concurrent drain check from calling `device.reset()` while a new op is still about to touch `device*`.
- Drain check after each decrement: both `_queued_reads` and `_queued_writes` must be zero.
- `_device_reset_done` CAS ensures exactly one queue thread calls `device.reset()` even with multiple queues racing to zero.
- `device` is a `shared_ptr` — `reset()` only destroys the object when all holders release (resync task must also drain first; that is a pre-existing concern, not introduced here).
- All counter RMWs, the `_shutting_down` store, and `all_idle()` loads are `seq_cst` to participate in a single total order, guaranteeing the increment-before-gate invariant on ARM and other weakly-ordered architectures.
- `probe_tick` is guarded by an explicit `if (!_shutting_down.load(seq_cst))` before capturing a local `shared_ptr` copy of `device`, preventing UAF if `device.reset()` fires between the load and the call. The probe timeout also stops re-arming after shutdown, preventing spurious queue-thread wakeups during drain.
- `wait_for_drain()` blocks on `_drain_complete.wait(false, acquire)`. Both `device.reset()` call sites (idle path in `begin_shutdown()` and non-idle path in `try_drain_device()`) store `true` with release ordering and call `notify_all()`, so the waiter unblocks in both cases. Returns immediately in the idle case if `begin_shutdown()` already called `device.reset()` synchronously.
- `ublkpp_queue_state::tgt` is now `shared_ptr<ublkpp_tgt_impl>` (was raw pointer). When `try_drain_device()` fires `notify_all()` and `wait_for_drain()` returns, the queue thread is still alive finishing `run_queue_loop` cleanup (`co_await scope.on_empty()`, `ublksrv_queue_deinit`). The shared_ptr keeps the impl alive through that entire window; without it there is a UAF between the `notify_all()` and the thread's final return.
- `~ublkpp_tgt_impl()` fires only when both `ublkpp_tgt._p` and every `qs->tgt` copy are released. When the last queue thread exits and its `qs` destructs, it triggers this destructor from inside `ublksrv_queue_handler`. At that point the thread is finishing but `queue_handlers[i].joinable()` is still true — joining from within the thread would deadlock, so `detach()` marks it as self-managed.
- Destructor fires `RELEASE_ASSERT` if `begin_shutdown()` was never called — crashing immediately at detection rather than proceeding into UAF.

## Fixes retained from previous iteration

- **`destroy()` missing `ctrl_dev = nullptr`**: prevents double-deinit if destructor runs after `destroy()`.
- **Unconditional `join()` in `destroy()`**: added `if (q.joinable())` guard.

## Testing

**Unit tests** — run via `conan build -s:h build_type=Debug --build missing .` on Linux (GCC, Ubuntu 24.04); all 14 tests pass:

- Pre-existing: `cqe_state.NextStateAllocatesDistinctStates`, `cqe_state.BuildCqeStateDataEncodesTargetBit`
- New `ShutdownDrain` suite (6 tests):
  - `AllIdleTrueInitially` — `all_idle()` returns true with both counters at zero
  - `AllIdleFalseWhenReadQueued` — false when `_queued_reads > 0`
  - `AllIdleFalseWhenWriteQueued` — false when `_queued_writes > 0`
  - `AllIdleFalseUntilBothCountersReachZero` — stays false until both counters reach zero
  - `BeginShutdownOnIdleSystemResetsDeviceSynchronously` — `begin_shutdown()` on an idle system calls `device.reset()` synchronously before returning; verified via a `TrackedDisk` that sets a flag in its destructor
  - `WaitForDrainReturnsImmediatelyAfterIdleShutdown` — `wait_for_drain()` returns within 200 ms after `begin_shutdown()` on an idle system (idle path sets `_drain_complete` before returning)

**Integration testing** (SIGTERM → `begin_shutdown()` + `wait_for_drain()` → verify `clean_unmount=1` on next mount, TSan under concurrent I/O) requires the NuBlox integration environment and is tracked separately.

### DIFF
--- a/src/target/ublkpp_tgt.cpp
+++ b/src/target/ublkpp_tgt.cpp
@@ -117,7 +117,14 @@ static exec::task< void > run_queue_loop(ublksrv_queue const* q, ublkpp_queue_st
                     // is_idle=false and preventing the probe from re-arming on subsequent fires.
                     ++probe_count;
                     if (cqe->res == -ETIME) {
-                        qs->tgt->device->probe_tick(q);
+                        // Gate check before capturing device: if _shutting_down is false,
+                        // begin_shutdown's seq_cst store has not committed yet — so
+                        // device.reset() has not been called. The local shared_ptr copy
+                        // keeps the object alive even if device.reset() fires immediately
+                        // after we capture it.
+                        if (!qs->tgt->_shutting_down.load(std::memory_order_seq_cst)) {
+                            if (auto dev = qs->tgt->device) dev->probe_tick(q);
+                        }
                         if (qs->is_idle) submit_probe_timeout(q);
                     }
                 } else {
@@ -288,21 +295,57 @@ static std::expected< std::filesystem::path, std::error_condition > start(std::s
     return res;
 }
 
+// Called after decrementing the in-flight counter during shutdown. If this is the last
+// in-flight op, wins the CAS and calls device.reset() to flush the backing store.
+// See all_idle() for the memory-ordering argument.
+static void try_drain_device(ublkpp_queue_state* qs) {
+    if (qs->tgt->metrics.all_idle()) {
+        bool expected = false;
+        if (qs->tgt->_device_reset_done.compare_exchange_strong(expected, true, std::memory_order_acq_rel)) {
+            TLOGI("All I/O drained after shutdown - flushing backing store")
+            qs->tgt->device.reset();
+        }
+    }
+}
+
 static exec::task< void > __handle_io_async(ublksrv_queue const* q, ublk_io_data const* data) {
     auto* qs = static_cast< ublkpp_queue_state* >(q->private_data);
 
-    auto device = reinterpret_cast< ublk_disk* >(q->dev->tgt.tgt_data);
     auto io = reinterpret_cast< async_io* >(data->private_data);
     io->_pool.clear();
     io->_tag = data->tag;
 
     auto const op = ublksrv_get_op(data->iod);
+
+    // Increment before the shutdown gate: a concurrent drain check that sees counters==0 would
+    // otherwise call device.reset() while we are still about to use the raw device* pointer.
+    // With the increment already in place, any concurrent drain check sees counter > 0.
+    // The seq_cst increment participates in the C++ total order S alongside begin_shutdown's
+    // seq_cst store and all_idle()'s seq_cst loads. Either the increment precedes the store in
+    // S (begin_shutdown's counter reads see it → skips reset) or the store precedes the
+    // increment in S (our gate check below sees _shutting_down=true → rejects).
     qs->tgt->metrics.record_queue_depth_change(q, op, true);
+
+    // Drain gate: reject reads/writes during shutdown before they reach the backing device.
+    // FLUSH is exempted: it completes instantly with result=0 and never dereferences device*.
+    // EAGAIN signals "temporarily unavailable, retry" so filesystems back off and retry once
+    // the new process reconnects the device (UBLK_F_USER_RECOVERY hot-restart path).
+    if (op != UBLK_IO_OP_FLUSH && qs->tgt->_shutting_down.load(std::memory_order_seq_cst)) {
+        qs->tgt->metrics.record_queue_depth_change(q, op, false);
+        TLOGD("Rejecting I/O [tag:{:#0x}] during shutdown", data->tag)
+        ublksrv_complete_io(q, data->tag, -EAGAIN);
+        // The rejected op may be the last in-flight — check drain here too.
+        // Without this, the common case (ops in-flight when begin_shutdown fires) would
+        // decrement to zero in this branch and never trigger device.reset().
+        try_drain_device(qs);
+        co_return;
+    }
 
     int result;
     if (op == UBLK_IO_OP_FLUSH) {
         result = 0;
     } else {
+        auto* device = reinterpret_cast< ublk_disk* >(q->dev->tgt.tgt_data);
         auto const* iod = data->iod;
         // Frame-local: io_uring reads iov contents at submit time (deferred to the queue loop's
         // submit_and_wait_timeout). thread_local would be overwritten by sibling __handle_io_async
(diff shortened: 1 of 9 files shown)


szmyd (commented):
**Critical: gate rejects with `-EIO` - defeats the hot-restart goal**

`__handle_io_async` rejects reads/writes during shutdown with:

```cpp
ublksrv_complete_io(q, data->tag, -EIO);
```

`-EIO` means "hard device failure - data may be lost." Filesystems treat it as a permanent error and go read-only or offline. At that point the careful bitmap flush and `clean_unmount=1` write are pointless - the filesystem is already in a bad state regardless.

The intent of `begin_shutdown()` is to support hot-restart: drain in-flight I/O, flush the backing store cleanly, exit, and let the new process reconnect via `UBLK_F_USER_RECOVERY`. For that to work, I/O that arrives during the drain window needs to signal "temporarily unavailable, retry" - not "dead device." That's `-EAGAIN`. The filesystem backs off and retries; once the new process is up and the device is live again, those I/Os succeed normally.

Fix:

```cpp
ublksrv_complete_io(q, data->tag, -EAGAIN);
```

The rest of the drain logic (counter management, CAS on `_device_reset_done`, FLUSH exemption) is unaffected.

---

**Nit: em dashes in log strings**

Two log messages use `—` (U+2014) instead of `-`:

```
"All I/O drained after shutdown — flushing backing store"
"No I/O in-flight at shutdown — flushing backing store immediately"
```

Use a hyphen in both.

src/target/ublkpp_tgt.cpp
```diff
@@ -520,12 +597,31 @@ void ublkpp_tgt_impl::destroy() {
...
+    // Safe only on process exit: detached threads still hold qs->tgt and will UAF if they
+    // re-enter the event loop after this destructor returns. The process must exit promptly.
+    for (auto& q : queue_handlers) {
+        if (q.joinable()) {
+            // If we are detaching without begin_shutdown(), the caller dropped the tgt without
+            // a graceful drain — backing store may be dirty and threads will UAF on the stale
+            // qs->tgt pointer if they re-enter the event loop after this destructor returns.
+            if (!_shutting_down.load(std::memory_order_relaxed))
+                TLOGE("ublkpp_tgt destroyed with live queue thread; begin_shutdown() was not called — backing store "
+                      "may be dirty")
```
szmyd: If this happens the process will UAF most likely leaving a confusing stack trace. I think a `RELEASE_ASSERT` may be better so we crash with a full stack trace at the moment of the detection?

szmyd (changes requested):
The two CAS tests are testing the stdlib, not your code. `DeviceResetDoneCasWinsOnFirstAttempt` and `DeviceResetDoneCasLosesOnSecondAttempt` just verify that `std::atomic<bool>::compare_exchange_strong` works correctly - that's not something you need to test. Replace them with a test that verifies begin_shutdown() on an idle system actually calls device.reset() synchronously, since that's the primary SIGTERM use case and is fully testable without the NuBlox environment.

src/target/ublkpp_tgt.cpp
```diff
@@ -520,12 +619,26 @@ void ublkpp_tgt_impl::destroy() {
...
+    // this destructor fires only when both ublkpp_tgt._p and every qs->tgt have been released.
+    // The most common case: the last queue thread to exit destructs qs, releasing its copy and
+    // triggering this destructor from inside ublksrv_queue_handler. At that point the thread
+    // is finishing but queue_handlers[i].joinable() is still true (not yet joined/detached).
+    // Joining from within the thread itself would deadlock; detach marks it as self-managed.
+    // The RELEASE_ASSERT ensures begin_shutdown() was called before this point.
+    for (auto& q : queue_handlers) {
+        if (q.joinable()) {
+            RELEASE_ASSERT(_shutting_down.load(std::memory_order_relaxed),
+                           "ublkpp_tgt destroyed with live queue thread; begin_shutdown() was not called");
```
Ilyin-Ilya, line 638: The destructor fires when the last shared_ptr<ublkpp_tgt_impl> ref drops to zero. Each queue thread holds qs->tgt for its entire lifetime. So the most common path is:

  1. Last queue thread finishes its work, destructs qs
  2. Dropping qs->tgt decrements the refcount to zero
  3. ~ublkpp_tgt_impl() runs from inside that queue thread
  4. q.join() on that thread → thread joining itself → deadlock

  detach() is the only option here: mark it self-managed so the thread can finish on its own after the destructor returns.

szmyd (approved):
Cool :+1:

PR #299, opened 2026-06-04
fix(raid1): self-heal read-determinism after both-present unclean shutdown (issue #290 phase 1)

Ilyin-Ilya:
## Background

RAID1 makes a weaker guarantee than Linux MD-RAID1 on unclean shutdown. When both legs are present and the array crashes while healthy (both legs at `clean_unmount=0`, `read_route=EITHER`, equal ages), writes in-flight at crash time may have landed on one leg but not the other. The array previously only logged a warning and continued serving reads round-robin across both legs.

This is a **read-determinism** bug, not a durability bug. The block-device contract — a given LBA always reads back the same bytes — is broken. Filesystems (XFS, etc.) and applications depend on read-determinism unconditionally, including for LBAs they never wrote. The same LBA can flip-flop between the two versions within a single mount as reads route A→B→A→B.

Closes #290 (Phase 1). Phase 2 (unified thin-aware resync) is tracked separately.

## What this PR does

Detects the both-present + unclean + `EITHER` + equal-age assembly case in `__init_bitmap_and_degraded_route` and self-heals:

```
Detection:  both legs present, clean_unmount=0, read_route=EITHER, equal ages
             ↓
Action:     bump canonical (device_a) age +16  → manufactures >1 gap vs stale on-disk age
            dirty whole bitmap                  → full canonical→stale resync scheduled
            read_route → DEVA                  → reads pinned to canonical
            device_b->unavail = true           → writes skip stale until probe_mirror clears

__become_active: skip writing device_b's SB (unavail guard)
             → canonical on-disk: age=N+16, route=DEVA
             → stale on-disk:     age=N,    route=EITHER  (gap preserved)
```

## Crash-mid-resync idempotency

On reassembly after a crash mid-resync, `pick_superblock` selects the canonical (higher age), `age_diff=16>1` sets `_device_b->new_device=true`, and the **existing** `new_device` path re-dirties the whole bitmap and resumes the full resync. No new superblock field; no new state machine. Idempotent across any number of crashes.

| Assembly | Path taken | Result |
|---|---|---|
| 1st (unclean, equal ages) | **new branch** | reads pinned to canonical, resync spawned |
| 2nd (crash mid-resync, age gap=16) | existing `new_device` path | bitmap re-dirtied, resync resumes |
| 3rd+ (if route=DEVA persists) | existing `unclean in degraded mode` path | bitmap re-dirtied, resync resumes |

## Changes

- **`raid1.cpp:__init_bitmap_and_degraded_route`**: replace the `else if (clean_unmount==0)` warning with the self-heal logic (+16 age, dirty all, pin to DEVA, mark device_b stale).
- **`raid1.cpp:__become_active`**: add `unavail` guard after `is_missing()` — skips writing the stale leg's SB to preserve the on-disk age gap for crash idempotency.
- **Tests**:
  - `UncleanShutdownWhileHealthy` → `UncleanShutdownBothPresentSelfHeal`: 3 writes to device_a (activate SB + bitmap sync + clean SB), 0 to device_b; `replica_states()` shows `device_a=CLEAN, device_b=ERROR, bytes_to_sync>0`.
  - New `UncleanBothPresentSelfHealIdempotentAfterCrash`: simulates on-disk state after crash mid-resync (canonical age=16, stale age=0), verifies second assembly uses existing `new_device` path (age=32, device_b gets `init_to`+SB).

## Acceptance criteria (from #290 Phase 1)

- [x] Both-present + `clean_unmount=0` + `read_route=EITHER` + equal-age assembly no longer serves diverged reads; reads pin to canonical until repair completes.
- [x] Detection manufactures a +16 age gap on the canonical leg only and dirties the whole bitmap, routing the rest through the existing age-divergent path.
- [x] Recovery is idempotent across repeated crashes (interrupt mid-recovery, reassemble via the age gap, recovery resumes and completes) — no new SB bit.
- [ ] On completion the ages re-equalize, the array returns to `read_route::EITHER`, and a clean shutdown sets `clean_unmount`. *(existing `__become_clean` machinery; covered by existing clean-degraded tests)*
- [x] Array is available for IO immediately on assembly (no blocking resync).
- [x] Tests: injected divergence on equal-age unclean assembly; crash-during-recovery resume; existing missing-leg / replacement paths unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.32.9] - 2026-06-04
+
+### Fixed
+
+- **RAID1 read-determinism after unclean shutdown (both legs present)**: when both legs were present at startup with equal ages, `clean_unmount=0`, and `read_route=EITHER`, the code previously only logged a warning. Writes in-flight at crash time may have landed on one leg but not the other, so round-robin reads of the same LBA could return different bytes on alternating accesses — a violation of the block-device read-determinism contract that filesystems (XFS, etc.) depend on unconditionally. The fix detects this case and self-heals: manufactures a +16 age gap on the canonical leg (device_a), pins reads to it via `read_route::DEVA`, dirties the whole bitmap, and marks device_b stale (`unavail`) so writes skip it until `probe_mirror` confirms physical health. `__become_active` skips persisting device_b's superblock (new `unavail` guard), preserving the on-disk age gap so any crash mid-resync reassembles through the existing `new_device` path rather than back into this branch. Idempotent across repeated crashes; no new superblock field. Addresses [issue #290](https://github.com/szmyd/ublkpp/issues/290) Phase 1.
+
 ## [0.32.8] - 2026-06-01
 
 ### Fixed
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=2.0"
 
 class UBlkPPConan(ConanFile):
     name = "ublkpp"
-    version = "0.32.8"
+    version = "0.32.9"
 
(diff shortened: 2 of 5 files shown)


szmyd (approved):
Looks correct!!

PR #301, opened 2026-06-05
fix(raid1): recover from resync-completed-before-stop race (issue #300)

Ilyin-Ilya:
## Problem

A race between the resync task and \`stop()\` in the destructor produced an on-disk state of \`DEVB + clean_unmount=1 + empty superbitmap\`. On second mount the degraded-clean invariant check threw \`std::runtime_error\`, making the volume permanently unassemblable. This caused the nublox \`resiliency-2-0-dense\` test failure tracked in issue #300.

**Root cause sequence:**
1. Array goes degraded. Route = DEVB. \`_sb->superbitmap_reserved\` has dirty bits set.
2. Health is restored; resync task launches and calls \`clean_region\` for each completed chunk, progressively zeroing \`_sb->superbitmap_reserved\`.
3. Resync calls \`__yield()\` (per-iteration sleep). At that moment \`stop()\` (from the destructor) CASes state → STOPPING. \`__yield()\` returns STOPPING → \`__run()\` returns STOPPING → the \`while(true)\` breaks **without reaching \`__become_clean\`**.
4. Destructor: \`sync_to\` sees an empty superbitmap → writes nothing. \`write_superblock\` copies \`_sb\` with zeros → writes \`DEVB + clean_unmount=1 + empty superbitmap\` to disk.
5. Second mount: \`route=DEVB, clean_unmount=1, superbitmap=zeros\` → invariant fires → throw → volume refuses to assemble.

## Fix

**Fix 1 — resync task** (\`_start()\` in \`raid1_resync_task.cpp\`):

Capture \`pages_before = dirty_pages()\` before each \`__run()\` call. When STOPPING fires, if \`pages_before > 0 && dirty_pages() == 0\` the resync cleared every dirty chunk during \`__run()\` but was interrupted in \`__yield()\` before \`complete()\` committed route→EITHER. Call \`complete()\` immediately so the destructor sees route=EITHER rather than leaving DEVA/DEVB + empty-superbitmap on disk. The \`pages_before > 0\` guard prevents a spurious \`complete()\` call for the Fix 2 recovery path (where \`dirty_region + load_from\` produces an empty bitmap with \`pages_before == 0\`). Calling \`complete()\` while \`stop()\` holds \`_launch_lock\` and waits at \`join()\` is safe — no I/O arrives during destruction, so \`dirty_region()\` cannot fire under \`_clean_transition_mutex\`, making \`complete()\` returning false impossible here.

**Fix 2 — mount invariant** (\`__init_bitmap_and_degraded_route\` in \`raid1.cpp\`):

Replace the throw with \`RLOGW + dirty_region(0, capacity())\`. The \`load_from\` call that follows reads the bitmap page from the active device; if it returns zeros (as the race-state disk would), \`isal_zero_detect\` clears the superbitmap bit — leaving the array in degraded mode for a clean transition on the next healthy shutdown. Defense-in-depth for any state not covered by Fix 1.

## Test

\`DegradedCleanEmptySuperbitmap\` in \`src/raid/raid1/tests/superblock/init_issues.cpp\` covers **Fix 2**:

- \`StrictMock<TestDisk>\` pair, 1 GiB each (→ exactly 1 bitmap page)
- \`disk_b\` carries the exact race-state SB: \`DEVB\`, \`age=1\` (wins \`pick_superblock\`), \`clean=1\`, empty superbitmap
- Constructor: Fix 2 fires → \`dirty_region(0, capacity())\` → \`load_from\` reads zeros → superbitmap bit cleared → array mounts in degraded mode without throwing
- Asserts: \`EXPECT_NO_THROW\`; route stays DEVB (resync disabled via \`toggle_resync(false)\`)

**Fix 1** (the \`pages_before\` guard in \`_start()\`) is covered by the nublox \`resiliency-2-0-dense\` integration test, which exercises the full resync-to-completion-then-stop sequence.

## Checklist

- [x] Fix 1: resync thread commits clean transition when stopped after emptying the bitmap
- [x] Fix 2: mount invariant replaced with warn + dirty-all fallback
- [x] Regression test exercises Fix 2 end-to-end (Fix 1 covered by integration test)
- [x] clang-format applied to modified files
- [x] CHANGELOG.md updated (0.32.10)
- [ ] CI — build must run on Linux (liburing / isa-l are Linux-only)

Closes #300

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-#Changelog
+# Changelog
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Changelog
+#Changelog
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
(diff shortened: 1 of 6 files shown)


szmyd (approved):
Got it; thanks! Looks good.

PR #305, opened 2026-06-06
fix(raid1): return -EAGAIN when __become_degraded SB write fails

Ilyin-Ilya:
## Summary

- **Root cause**: \`async_iov\` Site 1 returned the backup write result (SUCCESS) to the client even when \`__become_degraded\` failed to persist the degradation on disk. On a crash at that point both superblocks still showed \`route=EITHER\` at the old age; restart self-healed by resyncing device_a (old data) over device_b (new data), permanently destroying the acked write.
- **Fix**: when \`__become_degraded\` returns \`false\` (SB write failed), Site 1 returns \`-EAGAIN\` immediately — forces client retry rather than acking an unprotected write. \`_degraded_sb_pending\` is left \`true\` so the next I/O retries the SB write via \`__try_persist_degraded_sb\` before acking. Sites 2/3 follow the same pattern.
- **No inline Site-1 retry**: an earlier iteration added a second \`__become_degraded\` call at Site 1 to avoid a client round-trip on transient SB failure. Removed — the next I/O handles the retry identically through \`__try_persist_degraded_sb\`, and the extra call added complexity for a doubly-rare event (active fails AND the synchronous SB write fails).
- **\`-EAGAIN\` is correct (not \`-EIO\`)**: the data write to the surviving device succeeded; the SB write failure is transient and the already-degraded retry path closes the window on the next I/O.
- **All SB-fail sites use \`-EAGAIN\`**: async_iov Sites 1/2/3 and sync_iov Sites 2/3 (previously returned \`-EIO\` / \`std::errc::io_error\`).
- **\`__become_degraded\` simplified to \`bool\` and made idempotent**: every caller used the return only as a boolean. The already-degraded path now retries \`_degraded_sb_pending\` SB writes, re-capturing route state via \`__capture_route_state()\` to avoid stale \`cur_state\`.
- **Age increment NOT reverted on SB write failure**: any subsequent SB write must carry a higher age than the stale on-disk SB so \`pick_superblock\` selects the surviving device on restart.
- **\`_degraded_sb_pending\` visibility with \`nr_hw_queues > 1\`**: plain \`bool\` guarded by \`_ctrl_lock\`. Set to \`true\` pessimistically under that lock *before* \`write_superblock\` in \`__become_degraded\`, checked/cleared under the same lock in \`__try_persist_degraded_sb\`. Without this, a concurrent queue thread that loses the CAS and enters \`__try_persist_degraded_sb\` could read \`false\` while the flag store is still in-flight, prematurely acking an I/O. CAS and \`_degraded_sb_pending=true\` are a single atomic transaction under \`_ctrl_lock\`; \`__swap_device\` holds the same lock for its CAS, so both state-machine transitions are fully serialized.
- **\`was_pending\` snapshot guards double \`toggle_resync\`**: two coroutines may both pass the \`_degraded_sb_pending\` fast-path check before either clears the flag. The flag is cleared with a snapshot: only the coroutine that atomically reads \`true\` and stores \`false\` (under the lock) calls \`toggle_resync\`; a second concurrent \`launch()\` on an IDLE task would join the running thread and stall an I/O coroutine for the full resync duration.
- **\`failed_device->unavail.test_and_set\` hoisted**: runs unconditionally after the SB write branch, eliminating duplication.
- **Tests**: all SB-write \`Times()\` counts audited; deleted \`BecomeDegradedSbFailRetrySucceeds\` (tested removed code path); added \`SyncIoBecomeDegradedSbFailRetrySucceeds\` (active-fail sync Site 1, EAGAIN, then Site-2 retry succeeds); added \`SyncIoDegradedSbPersistRetryFails\` (sync Site 2 retry also fails); explicit \`Times(N)\` on all SB-write expectations to catch silent regressions.

Closes #302.

## Test plan

- [x] \`BecomeDegradedSbFail\`: Site 1 SB write fails → \`-EAGAIN\` (2 SB writes: initial fail + destructor)
- [x] \`WriteAndSbUpdateBothFail\`: Phase 1 \`-EAGAIN\`; Phase 2 degraded write succeeds (3 SB writes: Phase-1 fail, Phase-2 retry success, destructor)
- [x] \`DegradedSbPersistRetrySucceeds\`: Phase 1 \`-EAGAIN\`, Phase 2 normal ack
- [x] \`DegradedSbPersistRetryFails\`: Phase 1 and Phase 2 both \`-EAGAIN\`
- [x] \`SyncIoDegradedSbPersistRetryFails\`: sync Site 2 retry fails → \`resource_unavailable_try_again\`
- [x] \`SyncIoDegradedSbPersistRetrySucceeds\`: sync Site 2 retry succeeds → normal return
- [x] \`SyncIoBecomeDegradedSbFailRetrySucceeds\`: active-fail sync Site 1 → EAGAIN, then Site-2 retry succeeds
- [ ] CI: GccAddressSanitize, GccThreadSanitize, GccCoverage, ClangRelease all green


src/raid/raid1/raid1.cpp
```diff
@@ -897,12 +895,16 @@ disk_task< int > Raid1Disk::async_iov(ublksrv_queue const* q, ublk_io_data const
...
-        // would be wrong when disk_b received the write successfully.
+        // backup present. Always drain backup before returning: leaving it in-flight while the
+        // coroutine exits is unsafe.
         DEBUG_ASSERT(!become_degraded_ok || backup_task.has_value(),
                      "backup_task must exist when become_degraded succeeds"); // LCOV_EXCL_BR_LINE
         auto const backup_res = co_await *backup_task;
-        co_return backup_res >= 0 ? backup_res : -EAGAIN;
+        // Degradation not persisted on disk: a crash here would self-heal from the stale active
+        // device, overwriting whatever the backup wrote. Return EIO — the client must retry.
+        if (!become_degraded_ok) co_return -EIO;
```
Ilyin-Ilya: - PR #298 shutdown gate (EAGAIN): the device is being brought down cleanly for hot-restart. No data is in flight, nothing was written. "Temporarily unavailable — retry once the new process is up" is
  accurate.
  - PR #305 (EIO): the backup write succeeded (disk_b has new data), but __become_degraded couldn't persist the SB. On-disk state is now route=EITHER with disk_a holding stale data. A crash before the client
  retries triggers self-heal that resyncs disk_a over disk_b — permanently destroying the write. That's not "temporarily unavailable"; that's "I cannot guarantee durability of this write." EIO is the honest
  signal.

src/raid/raid1/raid1.cpp
```diff
@@ -730,32 +730,70 @@ bool Raid1Disk::__become_clean() {
...
+    std::shared_ptr< MirrorDevice > failed_device;
+    std::shared_ptr< ublk_disk > working_disk;
+    {
         std::lock_guard lock(_ctrl_lock);
-        auto age = _sb->fields.bitmap.age;
-        _sb->fields.bitmap.age = htobe64(be64toh(age) + 1);
-        return age;
-    }();
+        failed_device = failed_is_active ? cur_state->active_dev : cur_state->backup_dev;
+        working_disk = failed_is_active ? cur_state->backup_dev->disk : cur_state->active_dev->disk;
```
Ilyin-Ilya: The mutex, step by step
     
  There are two queue threads, T1 and T2, both completing I/Os on the same array at the same moment. Both see a failure and both call __become_degraded.

  The CAS narrows the race, but doesn't close it
  
  _read_route_cache.compare_exchange_strong(expected=EITHER, new_route=DEVA)

  This is atomic. Exactly one thread wins. Say T1 wins, T2 loses. T2 looks at the value it observed (old_route == DEVA == new_route) and concludes "someone else is handling degradation — I just need to make
  sure the SB is on disk" and calls __try_persist_degraded_sb.

  So far so good. But now there's a time gap. T1 won the CAS, but it hasn't stored _degraded_sb_pending = true yet — that happens inside the _ctrl_lock block a few instructions later. If T2 reads
  _degraded_sb_pending right now, it sees false and concludes "SB is already safe, nothing to do" — and returns true, letting the caller ack the write.

  Then T1 stores true, does the SB write, which fails. The write is acked to the client but the SB is never durable. Crash → data loss.

  Why seq_cst doesn't rescue you
  
  The C++ memory model guarantees that all seq_cst operations form a total order, but that order is not the same as wall-clock time. A valid interleaving is:

  [T1.CAS(seq_cst)]          → DEVA stored in cache line
  [T2.CAS_fail(seq_cst)]     → sees DEVA, loses
  [T2.load(pending) = false] → reads false ← PROBLEM
  [T1.store(pending = true)] → too late

  There is no rule in the standard that forces T1's store of pending to be sequenced-before T2's load of pending just because T1's CAS came first. They're different variables.

  Even x86 TSO only helps in practice because of how fast the lock prefix flushes the cache and how rarely the OS preempts T1 in that 10-instruction window — it's a probabilistic argument, not a guarantee.

  What the mutex gives you
  
  // T1, immediately after winning the CAS:
  {
      std::lock_guard lock(_ctrl_lock);   // acquire
      _degraded_sb_pending = true;        // store under lock
  }                                       // release — happens-before edge created

  // T2, in __try_persist_degraded_sb:
  {
      std::lock_guard lock(_ctrl_lock);   // acquire — must wait if T1 holds it
      if (!_degraded_sb_pending) return true;  // read under lock
  }

  A mutex unlock synchronizes-with the subsequent lock on the same object. That synchronizes-with edge establishes happens-before across threads: everything T1 did before the unlock (including the store of
  true) is visible to T2 after its lock.

  T2 has two possible executions:

  - T2 acquires the lock before T1: T2 reads false, returns early, goes off to do other work. T1 then acquires, stores true, does the SB write.
  - T2 tries to acquire while T1 holds it: T2 blocks. T1 stores true, releases. T2 unblocks, reads true, does the SB retry.

  In neither case can T2 read false while T1 is in the middle of setting it to true. The mutex collapses the gap between "T1 won the CAS" and "T1 stored the flag" into a single atomic critical section from
  T2's point of view.

  The full picture in one sentence

  The CAS ensures only one thread drives degradation; the mutex ensures any thread that checks _degraded_sb_pending either sees the flag already set, or waits until it is — so no thread can prematurely declare
  the SB safe and ack a write that isn't protected.

src/raid/raid1/raid1.cpp
```diff
@@ -730,32 +730,77 @@ bool Raid1Disk::__become_clean() {
...
         std::lock_guard lock(_ctrl_lock);
-        auto age = _sb->fields.bitmap.age;
-        _sb->fields.bitmap.age = htobe64(be64toh(age) + 1);
-        return age;
-    }();
+        if (_read_route_cache.compare_exchange_strong(old_route, new_route)) {
+            failed_device = failed_is_active ? cur_state->active_dev : cur_state->backup_dev;
+            working_disk = failed_is_active ? cur_state->backup_dev->disk : cur_state->active_dev->disk;
+            _sb->fields.bitmap.age = htobe64(be64toh(_sb->fields.bitmap.age) + 1);
+            _degraded_sb_pending = true;
```
Ilyin-Ilya, line 797: **Why `_ctrl_lock` here instead of an atomic?**

`_degraded_sb_pending` can't be a standalone atomic because it has to change in lockstep with two other operations that aren't individually atomic:

1. `_read_route_cache.compare_exchange_strong` (atomic, but its result must be visible together with 2 and 3)
2. `_sb->fields.bitmap.age` increment — plain `uint64_t` inside a packed struct; not atomically modifiable
3. `_degraded_sb_pending = true` — plain `bool`

**Without the lock, the race:**
- Thread A wins the CAS (route → DEVA), then is preempted before reaching this line.
- Thread B calls `__try_persist_degraded_sb`, acquires the lock, reads `_degraded_sb_pending == false`, and returns `true` — telling the caller the SB is durable.
- Thread A resumes, increments the age, sets the flag, and kicks off a `write_superblock` that may fail.

The caller has already been acked. On a crash before Thread A's write succeeds, both on-disk superblocks still carry `route == EITHER`, and restart self-heals from the wrong device.

Holding `_ctrl_lock` across all three steps closes this window: `__try_persist_degraded_sb` cannot observe `_degraded_sb_pending == false` while a concurrent winner is still in the middle of its transition.

szmyd (changes requested):
Still need to address above comment.

szmyd (approved):
Confirmed - the force-push adopts Option 3 (the `was_pending` snapshot on the winner path), and it closes the race correctly.

`_degraded_sb_pending` is armed `true` exactly once per episode (the CAS-winning branch, under `_ctrl_lock`) and can only be cleared by a serialized `{ was_pending = _degraded_sb_pending; _degraded_sb_pending = false; }` section. Both the winner (`__become_degraded`) and the loser (`__try_persist_degraded_sb`) now use that same read-and-clear, so across any number of queue threads exactly one caller observes the `true -> false` transition and only that one calls `toggle_resync`. This is independent of which `write_superblock` returns first or whether a thread is preempted between write and clear - so it's a proof, not the probabilistic mitigation Option 1 would have been.

Corners check out:
- `nr_hw_queues=1`: winner sets pending, writes, reads `was_pending==true`, toggles - resync still launches, no regression.
- Winner succeeds, loser cleared first: winner reads `was_pending==false`, skips the toggle; one launch.
- Winner's SB write fails but a concurrent loser's succeeds: winner skips the `if (sync_res)` block so `was_pending` stays false and it returns `-EAGAIN`; the loser persists and is the single toggle. The "unnecessary `-EAGAIN`" is documented and benign (client retries, next I/O acks).

The only residual is that the multi-queue interleave is reasoning-validated rather than unit-tested (it isn't deterministically reproducible single-threaded); the existing single-threaded retry tests plus the TSAN CI job are the backstop. Not a blocker.

LGTM.

PR #308, opened 2026-06-10
refactor(raid0): move sub_cmds from thread_local into caller frame

szmyd:
Replace thread_local std::array<StripeAccum,64> (with std::vector<iovec> per slot) with a caller-owned std::array<StripeAccum,64> where each slot holds a fixed std::array<iovec,16>. In async_iov the buffer lives in the coroutine frame (one heap allocation, stable for child lifetimes); in sync_iov it is a plain stack local.

The iovec bound k_max_iovecs_per_stripe=16 is max_io_size/min_stripe_size = 1 MiB / 64 KiB, matching the existing DEBUG_ASSERT_LE(nr_vecs,16u) in __failover_read_async.

Removes the two workarounds the thread_local hazard required:
- Eager io_uring_submit in async_iov (the queue loop's io_uring_submit_and_wait_timeout flushes all stripe SQEs when RAID0 suspends)
- iov_snap copy in Raid1Disk::__failover_read_async (iovecs now point into the RAID0 coroutine frame which outlives all child tasks)

Also removes DirtyGuard, _max_iovecs_per_stripe member, and the static_assert that existed solely to validate the dirty_mask bitmask width.


Ilyin-Ilya (commented):
1. No bounds check before the array write — potential OOB (medium-high)

  src/raid/raid0/raid0.cpp:233:
  acc.io_array[acc.nr_vecs++] = {buf_cursor, sz};
  The old vector grew dynamically; the new fixed array does not. k_max_iovecs_per_stripe = 16 is a hardcoded constant derived from max_io_size=1MiB / min_stripe_size=64KiB, but nothing in __distribute enforces
  this at runtime. If a configuration produces more than 16 iovecs per stripe (e.g. stripe_size < 64KiB or max_io_size > 1MiB), this silently corrupts adjacent frame memory.
  
  The old DEBUG_ASSERT_LE(nr_vecs, 16u) in __failover_read_async existed precisely to catch this, and is now removed with no replacement.

  Fix: Add a DEBUG_ASSERT_LT(acc.nr_vecs, k_max_iovecs_per_stripe) before the write, and add a constructor-time RELEASE_ASSERT in Raid0Disk that ceil(max_tx() / stripe_size) <= k_max_iovecs_per_stripe.

  ---
  2. sub_cmds block scope ends before the child tasks are awaited — technically UB (medium)
  
  In async_iov, sub_cmds is declared inside the } else { block, but the child tasks holding iovec pointers into sub_cmds.io_array are processed in the for loop outside that block:

  } else {
      std::array<StripeAccum, _max_stripe_cnt> sub_cmds{};   // ← declared here
      auto res = __distribute(sub_cmds, iovecs, addr, ...);  // stripe_tasks populated
      if (!res) co_return -EIO;
  }  // ← sub_cmds "destroyed" here

  for (auto& t : stripe_tasks) {                             // ← iovecs used here (co_await)

  In practice this works: StripeAccum is trivially destructible, the compiler won't call any destructor, and the coroutine frame storage isn't freed until the coroutine completes. But it's technically
  use-after-lifetime-end (UB), and a sufficiently aggressive compiler could reuse frame slots.

  The PR description says "iovecs now point into the RAID0 coroutine frame which outlives all child tasks" — that's true of the frame, but not of the block scope.

  Fix: Hoist sub_cmds to function scope (declare it before the if (DISCARD) { } else { } branch).

  ---
  3. Full 17 KB zero-init per I/O (low)
  
  64 × StripeAccum ≈ 17 KB. In sync_iov this is on the stack; in async_iov it's part of the coroutine frame allocation (one heap alloc, amortised). The old DirtyGuard only reset nr_vecs for the touched stripe
  slots (typically 2–4 × 4 bytes). For high-IOPS workloads this may be measurable. Not blocking, but worth a comment noting the tradeoff was accepted.

  ---
  4. Removal of failover_stale_iov.cpp without replacement (low)
  
  The test verified observable behavior: that __failover_read_async sees the correct iov_base/iov_len even when the caller mutates the original iov. The hazard is now structurally eliminated rather than
  papered over, so deleting the workaround-specific test is reasonable. However the invariant itself (failover reads from the right buffer) still deserves coverage, ideally via a simpler test that doesn't need
  to simulate a race.

Ilyin-Ilya (approved):
Good call, claude found some issues, the PR looks solid to me

PR #311, opened 2026-06-10
fix(raid0,raid1): DISCARD pool crash + RAID1 superblock version guard

Ilyin-Ilya:
## Summary

- **B1 (P1 crash) — RAID0 DISCARD pool exhaustion**: `prepare()` sized the CQE pool using the read/write fan-out formula `stripes_for_io(max_tx, stripe_size, N)` = `min(k, N)`. The DISCARD/WRITE_ZEROES path in `async_iov` always fans out to all N disks via `merged_subcmds`, consuming one `next_state()` pool slot per disk. For N > k (e.g., 10 disks, stripe_size=128 KiB, max_tx=512 KiB → k=5), `RELEASE_ASSERT_LT(_pool.size(), _pool.capacity())` fired on the (k+1)th disk's call — process crash on any file-backed RAID0 DISCARD.
- **E1 (latent) — RAID1 missing superblock version upper bound**: `load_superblock` had no guard against `version > SB_VERSION`, unlike RAID0's M3 check. A disk written by a future binary would be silently accepted; `__init_params`'s one-sided `if (sb_version < 2)` branch would compute `_reserved_size` using the v2 formula, placing all I/O at the wrong on-disk offset. Also exports `k_sb_version` from `raid1_superblock.hpp` (previously a file-local `SB_VERSION`).

### B1 — `src/raid/raid0/raid0.cpp`
- `prepare()`: remove the `stripes_for_io` k-bound; always accumulate `max_sqes_per_io` across all N children. `stripes_for_io` is still used in `async_iov` for `stripe_tasks.reserve()` — unaffected.
- New test: `Raid0PreparePoolSize.DiscardFanoutRequiresAllChildren` in `simple/discard.cpp` — asserts `prepare()` returns `max_sqes_per_io == N` for 10 mocked children each returning 1. Pre-fix this returned 5 (k-bounded); post-fix it returns 10.

### E1 — `src/raid/raid1/raid1_superblock.*`
- `raid1_superblock.hpp`: export `constexpr uint16_t k_sb_version = 2` in `ublkpp::raid1` namespace.
- `raid1_superblock.cpp`: remove file-local `SB_VERSION`; add version upper-bound guard in `load_superblock` after the UUID check — returns `errc::not_supported` for any `version > k_sb_version`. New devices have `version=0` (memset), so the guard is safe to place unconditionally.
- New test: `Raid1.FutureSuperblockVersionThrows` in `version_check.cpp` — mirrors the RAID0 `FutureSuperblockVersionThrows` pattern.

## Test plan

- [ ] CI: `GccAddressSanitize`, `GccThreadSanitize`, `GccCoverage`, `ClangRelease` all green
- [ ] `Raid0PreparePoolSize.DiscardFanoutRequiresAllChildren` passes — pool sized to N=10, not k=5
- [ ] `Raid1.FutureSuperblockVersionThrows` passes — constructor throws on version=3 superblock
- [ ] Existing RAID0 and RAID1 tests unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.32.13] - 2026-06-10
+
+### Fixed
+
+- **P1 (raid0): DISCARD/WRITE_ZEROES crashes file-backed arrays with more disks than the read/write stripe count**: `prepare()` sized the CQE pool using the read/write fan-out formula `stripes_for_io(max_tx, stripe_size, N)`, which caps at `k = min(ceil(max_tx/stripe_size)+1, N)`. The DISCARD path in `async_iov` always fans out to all N disks via `merged_subcmds`, consuming one pool slot per disk. For N > k (e.g., 10 disks, stripe_size=128 KiB, max_tx=512 KiB → k=5), `RELEASE_ASSERT_LT(_pool.size(), _pool.capacity())` fired on the (k+1)th disk's `next_state()` call — process crash. Fixed by removing the k-bound from `prepare()`: all N children contribute to `max_sqes_per_io`. Over-allocation by (N-k) slots in the read/write path is harmless.
+- **Latent (raid1): superblock with version > current not rejected at open**: `load_superblock` had no upper-bound version check, silently accepting a disk formatted by a future binary. `__init_params`'s one-sided `if (sb_version < 2)` branch would compute `_reserved_size` using the v2 formula, placing all I/O at the wrong on-disk offset on a hypothetical v3 disk. Fixed by adding the M3 guard (mirrors RAID0's existing check): any superblock with `version > k_sb_version` is now rejected with `errc::not_supported`. Also exports `k_sb_version` from `raid1_superblock.hpp` so tests can reference it.
+
 ## [0.32.12] - 2026-06-06
 
 ### Fixed
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=2.0"
 
 class UBlkPPConan(ConanFile):
     name = "ublkpp"
-    version = "0.32.12"
+    version = "0.32.13"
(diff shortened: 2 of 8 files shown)


szmyd (approved):
Reviewed both fixes in depth - they hold up.

**B1 (RAID0 DISCARD pool sizing)** - correct. `async_iov`'s DISCARD/WRITE_ZEROES path iterates `merged_subcmds`, which emits one coalesced region per touched disk, so a large discard fans out to all N children and each consumes a pool slot. `prepare()` sized the pool with `stripes_for_io` = `min(k, N)`, so for N > k the (k+1)th child tripped `RELEASE_ASSERT_LT(_pool.size(), _pool.capacity())`. Summing all N is the right fix, and the over-allocation in the read/write path (which touches at most k stripes) is harmless reserved capacity, never an invariant violation. Also correctly sizes RAID10 DISCARD fan-out (raid1 children report `max_sqes_per_io`=2). `stripes_for_io` is correctly retained for the `stripe_tasks.reserve()` hint in `async_iov`.

**E1 (RAID1 superblock version guard)** - correct, and the unconditional placement is safe. A fresh device is `memset` to all-zero (`version=0`) before the guard and only stamped to `k_sb_version` afterward, so `0 > 2` is false (new disks pass) while an existing v3 disk keeps its on-disk version and is rejected with `not_supported`. Mirrors RAID0's M3 guard exactly (same comparison, message, errc). Correctly does NOT add RAID0's symmetric upgrade branch - RAID1 v1/v2 have different `_reserved_size` layouts, so bumping the version field without migrating the layout would corrupt; leaving existing versions as-is (handled by `__init_params`) is right. No dangling `SB_VERSION` code references after the rename.

Both tests directly validate the fix (`DiscardFanoutRequiresAllChildren` asserts prepare returns N=10 not k=5; `FutureSuperblockVersionThrows` uses a StrictMock device_b to prove the constructor throws before touching the second leg).

Minor, non-blocking:
- The B1 test is prepare-level - it proves the pool is sized to N but doesn't push a full-stride DISCARD through `async_iov` end-to-end. Fair proxy for a unit test; worth a follow-up if there's an io_uring harness.
- Stale comment references to `SB_VERSION` remain in `version_check.cpp:60` and `raid0/.../init_issues.cpp:66` (comments only) - could rename to `k_sb_version` while here.

LGTM.

PR #312, opened 2026-06-10
fix(raid1): hold _clean_transition_mutex in Site-2 async_iov path (P0 data loss)

Ilyin-Ilya:
## Problem

`async_iov` Site 2 (backup-unavailable path) called `dirty_region()` +
`__become_degraded()` **without** holding `_clean_transition_mutex`,
unlike Sites 1 and 3 which always held it.

This opened a crash window that produced data loss against an acknowledged write.

### Race sequence

| Time | T1 (`__become_clean`) | T2 (`async_iov` Site 2) |
|---|---|---|
| t0 | acquires `_clean_transition_mutex` | — |
| t1 | CAS DEVA→EITHER | captures state DEVA (pre-CAS) |
| t2 | writes EITHER SBs to A and B (I/O in flight) | — |
| t3 | — | active write to A succeeds |
| t4 | — | `dirty_region() + __become_degraded()` (no mutex): CAS EITHER→DEVA, age N→N+1, writes **DEVA+N+1 to A**, ACKs caller |
| t5 | T1's in-flight EITHER writes (using `_sb` now at age N+1) land on disk, overwriting T2's DEVA write | — |
| t6 | releases mutex; H1 runs outside the mutex | — |
| crash | **process dies before H1 re-writes DEVA SBs** | — |

On restart: `pick_superblock` sees EITHER+N+1 on A and EITHER+N+1 on B.
Equal ages → tie-breaker selects A. Route = EITHER → round-robin reads.
B holds pre-T2 stale data → **reads from B return data that was never written** → data loss against the ACK at t4.

**Why H1 doesn't help:** H1 is defense-in-depth for the live-process case only. It runs after `_clean_transition_mutex` releases. Any crash between T1's second `write_superblock` completing and H1's first write bypasses it entirely.

## Fix

Wrap Site 2's `dirty_region() + __become_degraded()` in a `lock_guard` on `_clean_transition_mutex`, matching Sites 1 and 3:

```cpp
// Before
if (!backup_write) {
    _dirty_bitmap->dirty_region(addr, len);
    if (!__become_degraded(false, &state)) co_return -EAGAIN;
    co_return active_res;
}

// After
if (!backup_write) {
    bool const become_degraded_ok = [&] {
        std::lock_guard lock(_clean_transition_mutex);
        _dirty_bitmap->dirty_region(addr, len);
        return __become_degraded(false, &state);
    }();
    if (!become_degraded_ok) co_return -EAGAIN;
    co_return active_res;
}
```

With the mutex, T2 waits until T1 releases. T2's DEVA write always lands **after** T1's EITHER writes. On any crash:
- A = DEVA+N+1 (latest age)
- B = EITHER+N (older age)

`pick_superblock` selects A by age → route = DEVA → reads go to A only → **safe**.

**Lock-order safety:** `_clean_transition_mutex` → `_ctrl_lock` is the existing hierarchy (Sites 1 and 3 hold `_clean_transition_mutex` and call `__become_degraded` which acquires `_ctrl_lock` internally). Adding Site 2 to this order maintains the same hierarchy — no new deadlock risk.

**Coroutine safety:** `co_return` destroys RAII objects before final suspension, so the `lock_guard` in the Site 2 lambda is released before the coroutine is torn down.

## No new unit test

The race requires process-kill + restart semantics to observe: a test needs to stop the process mid-write and inspect raw on-disk superblock bytes. The ublksrv coroutine machinery cannot be driven from unit tests. The fix is a one-line lock application identical to the existing pattern in Sites 1 and 3 — the correctness argument is structural.

## Files changed

| File | Change |
|---|---|
| `src/raid/raid1/raid1.cpp` | Site 2: add `_clean_transition_mutex`; update `__become_clean` preamble and H1 comment |
| `src/raid/raid1/raid1_impl.hpp` | `_clean_transition_mutex` docblock: cover all three sites |
| `CHANGELOG.md` | New entry `[0.32.14]` |
| `conanfile.py` | Version bump 0.32.13 → 0.32.14 |

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **RAID1 (P0)**: Site 2 (`async_iov`/`sync_iov`, backup-unavail path) called `dirty_region()` and `__become_degraded()` without holding `_clean_transition_mutex`, unlike Sites 1 and 3.
 - A concurrent `__become_clean` could write EITHER superblocks after Site 2's DEVA write, leaving both devices at EITHER+equal-age on crash; `pick_superblock` tie-broke to round-robin, serving stale data from B against an acknowledged write.
-- Fixed by wrapping `__become_degraded()` in `_clean_transition_mutex` at all six failure sites (async + sync, Sites 1–3). `dirty_region()` moved lock-free before the mutex at all sites.
+- Fixed by wrapping `dirty_region()` + `__become_degraded()` together inside `_clean_transition_mutex` at all six failure sites (async + sync, Sites 1–3), making `dirty_pages()` a hard gate against premature clean transitions.
 
 ## [0.32.13] - 2026-06-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.32.14] - 2026-06-10
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


szmyd (dismissed):
Approving - this correctly closes the P0. Site 2 was missing the `_clean_transition_mutex` that Sites 1 and 3 already hold, which opened the crash window that loses an acknowledged write (both legs end up EITHER at equal age -> tie-break -> round-robin reads serve stale data from B). The fix is the right immediate action and matches the existing pattern, so let's land it.

Caveat / follow-up tracked in #313: the mutex now serializes essentially every write while degraded, not just the transition that causes degradation, so this is a correct stopgap rather than the structural fix. The lock exists to compensate for the on-disk superblock not durably reflecting acknowledged writes on crash. The structural alternative - use the existing superbitmap as a coarse (1 GiB/bit) write-intent log persisted before ack, which makes the EITHER+equal-age tie benign and lets the steady-state degraded write path stay lock-free - is captured in #313 with a decided Part 1 (the race-free snapshot primitive). Let's merge this now and investigate #313 as the replacement.

szmyd (approved):
looks good, will need to A/B test for performance against `latest-RelWithDebInfo` when we have a new `NubloxVerify` image.

PR #316, opened 2026-06-14
feat(metrics): add IO bytes/latency/error counters, resync gauges, and degraded-state gauge

Ilyin-Ilya:
## Summary

Adds six new Prometheus metrics to the ublkpp target and RAID-1 layers, covering throughput, latency, errors, resync progress, and degraded state. Together they close the main observability gaps in the UBD driver Sherlock dashboard.

### New metrics

| Metric | Type | Layer |
|---|---|---|
| `ublk_read_bytes_total` | Counter | Target (`entity=<volume_uuid>`) |
| `ublk_write_bytes_total` | Counter | Target (`entity=<volume_uuid>`) |
| `ublk_read_latency_us` | Histogram | Target (`entity=<volume_uuid>`) |
| `ublk_write_latency_us` | Histogram | Target (`entity=<volume_uuid>`) |
| `ublk_read_errors_total` | Counter | Target (`entity=<volume_uuid>`) |
| `ublk_write_errors_total` | Counter | Target (`entity=<volume_uuid>`) |
| `ublk_resync_remaining_kib` | Gauge | RAID-1 (`parent_id=<volume_uuid>`) |
| `ublk_resync_initial_kib` | Gauge | RAID-1 (`parent_id=<volume_uuid>`) |
| `ublk_raid_is_degraded` | Gauge | RAID-1 (`parent_id=<volume_uuid>`) |

### What can be displayed

**Throughput panel** — bytes per second per volume, split by direction:
```promql
rate(ublk_read_bytes_total{entity=~"$volume"}[5m])
rate(ublk_write_bytes_total{entity=~"$volume"}[5m])
```

**Latency panel** — P50 / P99 / P999 read and write latency per volume:
```promql
histogram_quantile(0.99, rate(ublk_read_latency_us_bucket{entity=~"$volume"}[5m]))
histogram_quantile(0.99, rate(ublk_write_latency_us_bucket{entity=~"$volume"}[5m]))
```

**Error rate panel** — read/write errors per second, useful for alerting on backing-device failures:
```promql
rate(ublk_read_errors_total{entity=~"$volume"}[5m])
rate(ublk_write_errors_total{entity=~"$volume"}[5m])
```

**Resync ETA panel** — estimated seconds until resync completes:
```promql
ublk_resync_remaining_kib{parent_id=~"$volume"}
  / rate(ublk_resync_progress_kib_sum{parent_id=~"$volume"}[5m])
```

**Resync progress panel** — percentage of resync complete (0–100):
```promql
(1 - (ublk_resync_remaining_kib{parent_id=~"$volume"}
       / ublk_resync_initial_kib{parent_id=~"$volume"})) * 100
```

**Resync throughput panel** — KiB/s currently being resynced:
```promql
rate(ublk_resync_progress_kib_sum{parent_id=~"$volume"}[5m])
```

**Degraded state panel** — 0/1 gauge per volume for alerting and time-in-degraded queries:
```promql
ublk_raid_is_degraded{parent_id=~"$volume"}

# Time spent degraded in last hour (seconds):
sum_over_time(ublk_raid_is_degraded{parent_id=~"$volume"}[1h]) * 15
```

## Implementation notes

- Latency is measured from just before `async_iov` to just after, covering both successful and failed IOs. FLUSH ops are excluded (they return 0 without touching the backing device).
- `ublk_resync_remaining_kib` is sourced from `dirty_data_est()` (the `_dirty_chunks_est × chunk_size` atomic estimate), updated after every resync sweep and yield. Both resync gauges reset to 0 on resync completion.
- `ublk_raid_is_degraded` is set to 1 in `__become_degraded` and back to 0 in `__become_clean` (only on the successful EITHER-route CAS path).

## Test plan

- [ ] `IOBytes.ReadBytesAccumulate` / `WriteBytesAccumulate` / `NonDataOpIgnored`
- [ ] `IOLatency.ReadLatencyObserved` / `WriteLatencyObserved` / `FlushOpIgnored`
- [ ] `IOError.ReadErrorAccumulates` / `WriteErrorAccumulates` / `FlushOpIgnored`
- [ ] All 14 existing tests continue to pass


src/raid/raid1/raid1.cpp
```diff
@@ -819,6 +823,7 @@ bool Raid1Disk::__become_degraded(bool failed_is_active, RouteState const* cur_s
         // LCOV_EXCL_START
         auto device_name = (new_route == read_route::DEVA) ? "device_b" : "device_a";
         _raid_metrics->record_device_degraded(device_name);
+        _raid_metrics->record_degraded_state(true);
```
szmyd, line 830: **`ublk_raid_is_degraded` misses arrays that boot already-degraded.**

This sets the gauge on the *runtime* `EITHER`→degraded transition, which is correct — but an array that **starts** degraded never reaches `__become_degraded`. `__init_bitmap_and_degraded_route()` stores the route directly in the missing-leg branch (`"...running with a missing device!"`) and the `"Raid1 is starting in degraded mode"` branch. Those arrays report `is_degraded=0` for their entire lifetime — including a permanently-missing leg — which defeats the alerting use case in the PR description.

`_raid_metrics` is already constructed (raid1.cpp:94) before `__init_bitmap_and_degraded_route()` (raid1.cpp:103), so reconciling the gauge there is cheap:

```cpp
// after route determination in __init_bitmap_and_degraded_route():
if (_raid_metrics)
    _raid_metrics->record_degraded_state(
        _read_route_cache.load(std::memory_order_acquire) != read_route::EITHER);
```

This also makes the healthy case explicit instead of relying on the gauge defaulting to 0.

src/target/ublkpp_tgt.cpp
```diff
@@ -366,13 +368,21 @@ static exec::task< void > __handle_io_async(ublksrv_queue const* q, ublk_io_data
         // coroutines spawned in the same CQE batch before the kernel sees the SQE. The coroutine
         // frame is alive across co_await, so the iov is valid through the whole IO lifetime.
         iovec iov{.iov_base = reinterpret_cast< void* >(iod->addr), .iov_len = iod->nr_sectors << SECTOR_SHIFT};
+        auto const io_start = std::chrono::steady_clock::now();
         result = co_await device->async_iov(q, data, &iov, 1, iod->start_sector << SECTOR_SHIFT);
+        auto const latency_us = static_cast< uint64_t >(
+            std::chrono::duration_cast< std::chrono::microseconds >(std::chrono::steady_clock::now() - io_start)
+                .count());
+        qs->tgt->metrics.record_io_latency(op, latency_us);
+        if (result >= 0) bytes_transferred = static_cast< uint32_t >(iov.iov_len);
```
szmyd, line 378: **Throughput counts requested length, not bytes actually transferred.** On a short/partial completion `result < iov.iov_len`, so this over-reports. Using `result` directly isn't clean either (some drivers `co_return 0` on inline success), so `iov.iov_len` is a defensible choice — worth a one-line comment noting the rationale and the partial-IO caveat.

szmyd (commented):
One inline note from a structured review (observability-only PR; no data-path concerns).

szmyd (approved):
LGTM - production-ready.

Verified the two prior review comments are addressed:
- Degraded-state gauge now covers boot-degraded arrays: record_degraded_state() is called at the end of __init_bitmap_and_degraded_route() after the route is settled, so missing-leg / replacement / unclean-degraded / start-degraded / both-present self-heal all report raid_is_degraded=1.
- iov_len-vs-result rationale is now documented at the byte-count site.

Also reviewed the bundled DISCARD/WRITE_ZEROES shutdown-drain UAF fix: _queued_other is now part of all_idle(), incremented before the shutdown gate and decremented symmetrically on both the reject and completion paths, same seq_cst protocol as reads/writes, and no device* is dereferenced after the decrement. ASan + TSan are green.

Non-blocking nit for a follow-up (not gating this approval): the _queued_other comment in src/metrics/ublk_io_metrics.hpp says 'record_queue_depth_change is a no-op for their op codes', but that function is exactly where _queued_other is maintained for op 3/5 (it is the mechanism this fix relies on). Reword to note these ops emit no depth histogram but are still tracked in _queued_other for the idle gate.

PR #317, opened 2026-06-17
perf(target): run queue handler threads under SCHED_FIFO at max priority

szmyd:
Queue threads call pthread_setschedparam(SCHED_FIFO, 99) on startup, placing them in the RT scheduling class and eliminating CFS jitter on the I/O hot path.

### DIFF
--- a/src/target/ublkpp_tgt.cpp
+++ b/src/target/ublkpp_tgt.cpp
@@ -37,7 +37,9 @@ SISL_OPTION_GROUP(ublkpp_tgt,
                    cxxopts::value< std::uint16_t >()->default_value("1"), "<queue_cnt>"),
                   (qdepth, "", "qdepth", "I/O Queue Depth per target",
                    cxxopts::value< std::uint16_t >()->default_value("128"), "<qd>"),
-                  (feature_zero_copy, "", "feature_zero_copy", "Enable ZeroCopy Feature", cxxopts::value< bool >(), ""))
+                  (feature_zero_copy, "", "feature_zero_copy", "Enable ZeroCopy Feature", cxxopts::value< bool >(), ""),
+                  (sched, "", "sched", "Queue thread scheduler policy (other, fifo)",
+                   cxxopts::value< std::string >()->default_value("fifo"), "<policy>"))
 
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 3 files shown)


src/target/ublkpp_tgt.cpp
```diff
@@ -159,6 +161,8 @@ static exec::task< void > run_queue_loop(ublksrv_queue const* q, ublkpp_queue_st
 
 static void* ublksrv_queue_handler(std::shared_ptr< ublkpp_tgt_impl > target, int q_id, sem_t* queue_sem,
                                    int* queue_ok) {
+    sched_param sp{.sched_priority = sched_get_priority_max(SCHED_FIFO)};
```
Ilyin-Ilya: 1. Return value of pthread_setschedparam is silently ignored

  This is the main concern. pthread_setschedparam fails with EPERM when the process lacks CAP_SYS_NICE or RLIMIT_RTPRIO is 0, which is the default in most container environments and unprivileged Linux
  installs. The code proceeds silently and runs at normal priority, with no indication that the intended optimization didn't take effect.

  // Current — failure is silent
  pthread_setschedparam(pthread_self(), SCHED_FIFO, &sp);

  // Better — at minimum, log on failure
  if (int rc = pthread_setschedparam(pthread_self(), SCHED_FIFO, &sp); rc != 0) {
      RLOGW("Failed to set SCHED_FIFO on queue thread {}: {} (running at normal priority)", q_id, strerror(rc));
  }

  2. sched_get_priority_max(SCHED_FIFO) can return -1 on error

  Unlikely in practice, but technically sched_get_priority_max returns -1 on failure and errno is set. Worth at least asserting or using the literal 99 (which is hardcoded in the PR description anyway).

  3. Priority 99 is aggressive — starves everything else on that CPU if the queue thread spins or blocks in a tight loop

  SCHED_FIFO at priority 99 is the highest possible RT priority. If a queue thread ever gets stuck in a busy-wait or a slow path (e.g., a slow disk backend, a lock contention path that doesn't yield), it will
  starve all other threads on that core, including the kernel's own softirqs on some kernels. A lower RT priority (e.g., 50 or max - 1) would still preempt all SCHED_OTHER work while leaving headroom for
  critical kernel threads.

szmyd: This is all true; however the queue threads are non-blocking, all I/O is done asynchronously so there will be no unyielding system calls (other than writing a degraded SuperBlock). The fact that we have to expose privileged devices through the kernel means there will be an error about privileges _long_ before we reach this point where we set our scheduler (i.e. registering the ublk control device). Can add error check, but likely will be a no-op and untestable.

Ilyin-Ilya (approved):

PR #330, opened 2026-07-21
feat(api): memory estimation for RAID1 and target

szmyd:
## Summary

Adds a lightweight, options-aware memory estimation API so callers can budget
memory before provisioning a volume.

### New API

```cpp
// include/ublkpp/raid.hpp
namespace ublkpp::raid1 {
    // Worst-case device memory: one superblock + all bitmap pages dirty.
    // Uses chunk_size from SISL options. Does NOT include queue overhead.
    uint64_t estimate_device_overhead(uint64_t volume_size) noexcept;
}

// include/ublkpp/target.hpp
struct ublkpp_tgt {
    // ublksrv I/O buffers + queue thread stacks, shared across all devices.
    // Uses qdepth / max_io_size / nr_hw_queues from SISL options.
    static uint64_t estimate_queue_memory() noexcept;
};
```

### Design notes

- **RAID0** has no device-specific overhead worth tracking (no bitmap, no dirty
  state); the function was removed after evaluation.
- **RAID1** worst-case is all bitmap pages dirty: one superblock (4 KiB) +
  `num_pages × sizeof(Bitmap::PageData)` (24 bytes each) + one shared clean
  page (4 KiB) + `num_pages × 4 KiB` for dirty page buffers. At 32 KiB chunk
  size each page covers 1 GiB, so a 2 TiB volume is ~8 MiB worst-case.
- `PageData` size is guarded by a `static_assert` so a struct layout change
  breaks the build rather than silently producing wrong estimates.
- Both functions read live SISL runtime options, so estimates match the actual
  configuration of a running target.

## Test plan

- `Raid1MemoryEstimation.*` - exact-byte assertions at 512 MiB, 1 TiB, 10 TiB,
  and 100 TiB; plus a linearity ratio check.
- `MemoryEstimation::QueueMemory` - range check against default SISL options
  (70-75 MiB expected with qdepth=128, max_io_size=512 KiB, nr_hw_queues=1).
- All 14 tests pass clean under ASAN.


CHANGELOG.md
```diff
@@ -255,6 +264,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 0.20.x
 - raid1: Fix on_io_complete metrics mapping after device swap - I/O metrics were incorrectly attributed to the wrong device when completions occurred after swap_device() changed the route.
 - Fix some more cases where __capture_route_state was being misused.
+>>>>>>> main
```
Ilyin-Ilya: **P1 — unresolved merge conflict marker**

This `>>>>>>> main` line was left in from a merge commit and must be removed before merge.

include/ublkpp/lib/memory_constants.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+#include <cstdint>
+
+namespace ublkpp {
+
+// OS thread stack size (Linux default)
+constexpr uint64_t k_thread_stack_size = 8 * 1024 * 1024; // 8 MiB
+
+// Typical async_io structure size (from target/ublkpp_tgt.cpp)
+constexpr uint64_t k_async_io_size = 40; // ~40 bytes
```
Ilyin-Ilya: **P1 — wrong constant, no guard**

`sizeof(async_io) == 32`, not 40. The struct is:
```cpp
struct async_io {
    std::vector<cqe_state> _pool{};  // 24 bytes
    int _tag{-1};                    // 4 bytes + 4 bytes padding
};
```

The PR correctly guards `Bitmap::PageData` with a `static_assert`; `async_io` should get the same treatment:
```cpp
static_assert(sizeof(async_io) == 32, "async_io size changed — update k_async_io_size");
constexpr uint64_t k_async_io_size = sizeof(async_io);  // exact, not approximate
```

Note: `ublkpp_tgt.cpp:434` already uses `sizeof(struct async_io)` directly for the ublksrv slot size — the constant should match.

src/raid/raid1/raid1.cpp
```diff
@@ -1208,6 +1209,19 @@ std::pair< std::shared_ptr< ublk_disk >, std::shared_ptr< ublk_disk > > replicas
     return r1->replicas();
 }
 
+uint64_t estimate_device_overhead(uint64_t volume_size) noexcept {
+    auto const chunk_size = SISL_OPTIONS["chunk_size"].as< uint32_t >();
+    constexpr uint64_t bits_per_byte = 8;
+    uint64_t page_width = static_cast< uint64_t >(chunk_size) * k_page_size * bits_per_byte;
+    uint64_t num_pages = (volume_size / page_width) + ((volume_size % page_width) ? 1 : 0);
```
Ilyin-Ilya, line 1217: **P2 — division by zero if `chunk_size == 0`**

If `chunk_size` is 0 (malformed option or called before SISL_OPTIONS_LOAD), `page_width == 0` and `volume_size / page_width` is UB. The guard in `MirrorDevice`'s constructor (`k_min_chunk_size > chunk_size → throw`) doesn't protect this standalone estimation function.

Suggested fix:
```cpp
if (chunk_size < k_min_chunk_size) return 0;  // or assert/document precondition
```

docs/API.md
```diff
@@ -0,0 +1,772 @@
...
+**Throws:** `std::invalid_argument` if constraints violated
+
+**Constraints:**
+- Both devices must have same capacity
+- Both devices must have `direct_io = true`
+
+#### Methods
+
+```cpp
+static uint64_t estimate_device_overhead(uint64_t volume_size) noexcept;
```
Ilyin-Ilya: **P2 — wrong API shape: `Raid1Disk` static method doesn't exist**

The public header (`include/ublkpp/raid.hpp`) declares a free function:
```cpp
namespace ublkpp::raid1 {
    uint64_t estimate_device_overhead(uint64_t volume_size) noexcept;
}
```

`Raid1Disk` is an internal class not exposed in the public API — per project convention, drivers and RAID arrays are opaque `disk_handle`s from factory functions. All call sites in these docs that read `Raid1Disk::estimate_device_overhead(...)` should be `ublkpp::raid1::estimate_device_overhead(...)`.

docs/INTEGRATION.md
```diff
@@ -0,0 +1,479 @@
...
+
+## Conan Integration
+
+### Basic conanfile.txt
+
+Create a `conanfile.txt` in your project:
+
+```ini
+[requires]
+ublkpp/0.21.0@oss/main
```
Ilyin-Ilya: **P2 — stale version (4 occurrences)**

This PR bumps the version to `0.36.0`, but `ublkpp/0.21.0@oss/main` appears in INTEGRATION.md at lines 20, 36, 62, and 276. All four should be `ublkpp/0.36.0@oss/main`.

docs/INTEGRATION.md
```diff
@@ -0,0 +1,479 @@
...
+    // Create backend devices
+    auto disk_a = std::make_shared<ublkpp::FSDisk>(argv[1]);
+    auto disk_b = std::make_shared<ublkpp::FSDisk>(argv[2]);
+
+    std::cout << "Device A: " << disk_a->capacity() << " bytes\n";
+    std::cout << "Device B: " << disk_b->capacity() << " bytes\n";
+
+    // Create RAID1 mirror
+    auto uuid = boost::uuids::random_generator()();
+    auto raid = std::make_shared<ublkpp::Raid1Disk>(uuid, disk_a, disk_b);
```
Ilyin-Ilya: **P2 — internal class exposed in docs**

`std::make_shared<ublkpp::Raid1Disk>(...)` uses an internal implementation class. Per project convention (CLAUDE.md): *"Drivers and RAID arrays are not exposed as classes; they are opaque `disk_handle`s constructed via factories."*

Expected usage:
```cpp
auto disk_a = ublkpp::make_fs_disk(argv[1]);
auto disk_b = ublkpp::make_fs_disk(argv[2]);
auto raid   = ublkpp::make_raid1_disk(uuid, disk_a, disk_b);
```

Same issue appears throughout INTEGRATION.md, LIBRARY.md, and EXTENDING.md with `FSDisk`, `Raid1Disk`, etc.

docs/INTEGRATION.md
```diff
@@ -0,0 +1,479 @@
...
+**Environment Variables:**
+```bash
+export UBLK_NR_HW_QUEUES=4
+export UBLK_QUEUE_DEPTH=256
+export UBLK_MAX_IO_SIZE=1048576  # 1 MiB
+```
+
+**Command Line:**
+```bash
+./my_app --nr_hw_queues 4 --queue_depth 256 --max_io_size 1048576
```
Ilyin-Ilya: **P1 — wrong CLI flag: `--queue_depth` is not a registered SISL option**

The SISL option is registered as `qdepth` in `ublkpp_tgt.cpp`:
```cpp
SISL_OPTION_GROUP(ublkpp_tgt,
    (qdepth, "", "qdepth", "Queue depth", cxxopts::value<uint16_t>()->default_value("128"), "")
```

Passing `--queue_depth 256` on the command line would silently be ignored and the default of 128 would be used instead. The correct flag is `--qdepth`. Same issue on line 384 (`queue_depth` in the option registration snippet).

docs/EXTENDING.md
```diff
@@ -0,0 +1,712 @@
...
+                        ublk_io_data const* data,
+                        sub_cmd_t sub_cmd,
+                        iovec* iovecs,
+                        uint32_t nr_vecs,
+                        uint64_t addr) override {
+        auto op = ublksrv_queue_get_op(q, data);
+
+        if (op == UBLK_IO_OP_READ) {
+            // Async S3 GetObject with range
+            auto range = fmt::format("bytes={}-{}", addr, addr + iov_len - 1);
```
Ilyin-Ilya: **P2 — `iov_len` is undefined in scope**

`iov_len` is not declared anywhere in this function. The function receives `iovec* iovecs, uint32_t nr_vecs` — the byte length needs to be computed from the array, for example:
```cpp
uint64_t total_len = 0;
for (uint32_t i = 0; i < nr_vecs; ++i) total_len += iovecs[i].iov_len;
auto range = fmt::format("bytes={}-{}", addr, addr + total_len - 1);
```

Ilyin-Ilya (changes requested):
## Review summary

Overall: **request changes** — 3 P1 issues, 5 P2 issues.

### P1 — High
1. Conflict marker in `CHANGELOG.md` (inline)
2. `k_async_io_size = 40` is wrong — `sizeof(async_io) == 32` (inline)
3. `--queue_depth` in docs doesn't match the real SISL key `--qdepth` — breaks user CLI usage (inline)

### P2 — Medium
4. Division by zero if `chunk_size == 0` in `estimate_device_overhead` (inline)
5. `Raid1Disk::estimate_device_overhead` doesn't exist — public API is `ublkpp::raid1::estimate_device_overhead()` free function (inline)
6. Four stale `ublkpp/0.21.0@oss/main` version refs in INTEGRATION.md — PR bumps to 0.36.0 (inline)
7. Internal classes (`Raid1Disk`, `FSDisk`) exposed in docs — convention is opaque factory functions `make_raid1_disk`, `make_fs_disk` (inline)
8. `iov_len` is undefined in scope in EXTENDING.md (inline)

### P3 — Low
9. `Bitmap::PageData` size is protected by a `static_assert`; `async_io` (the other hardcoded size) is not.

### Commit messages
Three commits have vague messages that should be more descriptive:
```
❌ "Added some comments regarding missed uts."
✅ "chore(raid1): note missing test coverage for removed lock-free bitmap path"

❌ "Better docs"
✅ "docs: expand library integration, extension, and API reference guides"

❌ "Raid0 has no memory overhead."
✅ "docs(api): remove RAID0 memory estimation — no device-specific overhead to track"
```

docs/API.md
```diff
@@ -0,0 +1,773 @@
+# API Reference
+
+Complete API documentation for ublkpp library.
+
+## Table of Contents
+
+- [Core Types](#core-types)
+- [UblkDisk Base Class](#ublkdisk-base-class)
```
Ilyin-Ilya: **P2 — `UblkDisk` is not the real class name (40+ occurrences in this file and EXTENDING.md, LIBRARY.md)**

The actual public base class in `include/ublkpp/` is `ublk_disk` (lower_snake_case), consistent with the project naming convention for public API types. `UblkDisk` does not exist — any user who copies an example from these docs will get a compile error.

Global replace throughout all three doc files:
```
UblkDisk → ublk_disk
std::shared_ptr<UblkDisk> → ublk_disk::disk_handle  (or disk_handle if the alias is in scope)
class RAMDisk : public ublkpp::UblkDisk → class RAMDisk : public ublkpp::ublk_disk
```
Also affects `LIBRARY.md` (first occurrence: line 63) and `EXTENDING.md` (first occurrence: line 3).

docs/API.md
```diff
@@ -0,0 +1,773 @@
...
+- `device_id`: Specific device ID, or -1 for auto-assign
+
+**Returns:**
+- `std::unique_ptr<ublkpp_tgt>` on success
+- `std::unexpected(error)` on failure
+
+**Example:**
+```cpp
+auto uuid = boost::uuids::random_generator()();
+auto disk = std::make_shared<FSDisk>("/dev/sda");
```
Ilyin-Ilya: **P2 — internal class `FSDisk` still exposed here (missed when fixing other files)**

The fix was applied to INTEGRATION.md, LIBRARY.md, and EXTENDING.md, but this instance in API.md was missed. Per project convention, `FSDisk` is an internal class; the public API is the factory:
```cpp
auto disk = ublkpp::make_fs_disk("/dev/sda");
```

docs/API.md
```diff
@@ -0,0 +1,773 @@
...
+
+```cpp
+static uint64_t estimate_queue_memory() noexcept;
+```
+
+Estimate target queue memory overhead.
+
+**Returns:** Estimated memory in bytes based on SISL options:
+- `nr_hw_queues` (default: 1)
+- `queue_depth` (default: 128)
```
Ilyin-Ilya: **P2 — `queue_depth` is not a SISL option key**

The SISL option registered in `ublkpp_tgt.cpp` is `qdepth`, not `queue_depth`. A user reading this list who tries `SISL_OPTIONS["queue_depth"]` would get a runtime throw. Should read:
```
- `qdepth` (default: 128)
```

docs/LIBRARY.md
```diff
@@ -0,0 +1,597 @@
...
+## Memory Estimation API
+
+ublkpp provides APIs to estimate memory consumption before device creation:
+
+### Target Queue Memory
+
+```cpp
+// Estimate per-target overhead (I/O queues, threads)
+uint64_t queue_mem = ublkpp_tgt::estimate_queue_memory();
+// Uses SISL options: nr_hw_queues, queue_depth, max_io_size
```
Ilyin-Ilya: **P2 — `queue_depth` option name wrong**

Same as API.md:668 — the SISL key is `qdepth`, not `queue_depth`. Should read:
```cpp
// Uses SISL options: nr_hw_queues, qdepth, max_io_size
```

Ilyin-Ilya (changes requested):
## Re-review after 2026-07-21 fixes

Thank you for the quick turnaround — all 8 original findings were addressed:

✅ Conflict marker removed from CHANGELOG.md  
✅ `k_async_io_size = 40` → `sizeof(async_io)` inline (exact, can't drift)  
✅ `--queue_depth` → `--qdepth` in INTEGRATION.md  
✅ Division-by-zero guard added (`chunk_size < k_min_chunk_size → return 0`)  
✅ `Raid1Disk::estimate_device_overhead` → `ublkpp::raid1::estimate_device_overhead` free function  
✅ Stale `0.21.0` → `0.36.0` (all 4 occurrences)  
✅ `make_shared<Raid1Disk/FSDisk>` → factory functions in INTEGRATION.md, LIBRARY.md, EXTENDING.md  
✅ `iov_len` undefined → computed from iovecs array  

Three issues remain, all in docs:

### P2 — `UblkDisk` is not the real public base class name (40+ occurrences across 3 files)

All three new doc files use `UblkDisk` (PascalCase), but the actual public API base class is `ublk_disk` (lower_snake_case), per the project naming convention: *"Public API types use `lower_snake_case`."* Every example that reads `class RAMDisk : public ublkpp::UblkDisk` fails to compile — the class is `ublkpp::ublk_disk`. This affects API.md, EXTENDING.md, and LIBRARY.md.

### P2 — `make_shared<FSDisk>` missed in API.md line 620

The factory-function fix was applied to INTEGRATION.md, LIBRARY.md, and EXTENDING.md, but one instance in API.md was missed (inline).

### P2 — `queue_depth` option name still wrong in 2 places

API.md:668 and LIBRARY.md:508 still name the SISL option `queue_depth` instead of `qdepth` (inline on each).

src/target/ublkpp_tgt.cpp
```diff
@@ -655,4 +656,22 @@ ublkpp_tgt_impl::~ublkpp_tgt_impl() {
...
 }
 
+uint64_t ublkpp_tgt::estimate_queue_memory() noexcept {
+    // Read current runtime configuration
+    auto const qdepth = SISL_OPTIONS["qdepth"].as< uint16_t >();
+    auto const max_io_size = SISL_OPTIONS["max_io_size"].as< uint32_t >();
+    auto const nr_hw_queues = SISL_OPTIONS["nr_hw_queues"].as< uint16_t >();
+
+    uint64_t ublksrv_buffers = static_cast< uint64_t >(nr_hw_queues) *
+        ((static_cast< uint64_t >(qdepth) * max_io_size) + (qdepth * sizeof(async_io)));
```
Ilyin-Ilya, line 666: do we need to add something like
nr_hw_queues * qdepth * max_sqes_per_io * sizeof(cqe_state)

Currently we don't count cqe pool, right?

szmyd: You're right, we don't. The pool is `max_sqes_per_io × sizeof(cqe_state)` per slot — topology-dependent (1 for a single disk, 2 for RAID1, up to stripe-count for RAID0) and not available without knowing the disk stack. At typical values it's <10 KiB total against 64+ MiB of I/O buffers, so I've left it out but added a comment documenting the omission and why (86804b2).

Ilyin-Ilya (approved):

